PR #314, opened 2022-04-18
feat(sdk): accept JWTs in SDK init options map

### LINKED ISSUE #312: feat(sdk): Store JWTs in local storage
# Why

Persist login across refreshes, new tabs etc.

# How

Local storage with CORS

### LINKED ISSUE #329: BUG([sdk]): ddt wallet regression
## I'm submitting a ...

- [x] bug report
- [ ] feature request

## What is the current behavior?
ddt does not show wallet in ddt wallet ls. it also fails when running ddt p2p auth 

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem
`$ ddt wallet init walletname 
(works) 
$ ddt wallet ls 
(no output) 
$ ddt p2p auth -p 8787 0x45a5e14eaf2f505067df379a6f527113b35afc0d -w testing
Enter your password:  ****

Error: ENOENT: no such file or directory, open '/home/billk/.local/share/ddt/wallets/testing'
    at Object.openSync (node:fs:585:3)
...
`
## What is the expected behavior?
should find the wallet in the correct directory 

## Please tell us about your environment:

Ubuntu 20.04.4 LTS

Version:
Browser or OS:

crimeminister:
# Description

Update the SDK `init` function to accept JWTs in the initialization options map. Provide a corresponding API call for retrieving the options map that will allow the SDK to be re-initialized with the same state.

Closes #312
Closes #329 

- [x] decoded JWTs passed in as options and store in "vault"
- [x] add `options` call to retrieve SDK initialization options
- [x] update `ddt sdk core authenticate` to take initial JWT arguments
- [x] add `ddt sdk options` command
- [x] collapse p2p config to have single address
- [x] split IPFS config to have separate r/w addresses
- [x] update IPFS client init to return a promise
  - separate r/w paths mean that we may need/want to make a call to each one during initialization
  - these calls return node information that may be pertinent, e.g. Go vs JS flavor (which have different semantics for some commands)
  - returned promise resolves after any such calls complete
- [x] fix `ddt wallet ls` command to use new wallet location
- [x] fix `ddt wallet rm` command to use new wallet location
- [x] add utility `to-map` command to process CLI args into standard form
- [x] various `sdk.v1.init` options changes
  - log level is now supplied as `:log/level` / `"log/level"`
  - accept `credential/jwt` key in SDK options map

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/src/main/com/kubelt/lib/init.cljc
+++ b/src/main/com/kubelt/lib/init.cljc
@@ -6,9 +6,10 @@
    [taoensso.timbre :as log])
   (:require
    [com.kubelt.ipfs.client :as ipfs.client]
-   [com.kubelt.lib.detect :as detect]
+   [com.kubelt.lib.jwt :as lib.jwt]
    [com.kubelt.lib.multiaddr :as lib.multiaddr]
    [com.kubelt.lib.util :as util]
+   [com.kubelt.lib.vault :as lib.vault]
    [com.kubelt.lib.wallet :as lib.wallet]
    [com.kubelt.proto.http :as proto.http])
   (:require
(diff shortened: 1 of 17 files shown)


src/main/com/kubelt/lib/init.cljc
```diff
@@ -220,12 +285,28 @@
         ;; can detect a locally-running p2p node.
         default (:client/p2p system)
```
tangrammer, line 286: why not use `(:client/p2p options)` instead of local def `system` ?

tangrammer: also good to rename default to something more meaningful

tangrammer: so, understanding it now ... trying to figure out a better way of passing optional values here ...

crimeminister: Good eye; this bit of broken stuff is a holdover from when the backend stuff worked a little differently. Will have a chat to get some insight into how things are expected to work in the future and perhaps we can fix in a subsequent PR.

crimeminister: Have removed this and reworked some of the related code.

src/main/com/kubelt/lib/init.cljc
```diff
@@ -68,27 +141,37 @@
...
+                     :read/scheme read-scheme
+                     :read/host read-host
+                     :read/port read-port
+                     :write/scheme write-scheme
+                     :write/host write-host
+                     :write/port write-port
+                     ;; Set a global timeout for *all* requests:
+                     ;;:client/timeout 5000
+                     }]
         (ipfs.client/init options))
```
crimeminister, line 175: @tangrammer The design issue I wanted to discuss is that this function call now returns a promise. Ideally, integrant would resolve this promise before merging the result into the system map. A little more research is required to determine whether or not integrant supports that capability, or can be made to support it without too much fuss. Alternately, if one of the other libraries in this area has this feature it might be worth looking at; I have recently heard good things about `donut-power/system`, for example. Ideally whatever library we choose in this area would support Clojure futures in a similar fashion.

drew-patel (dismissed):

drew-patel (approved):

PR #324, opened 2022-04-20
[proposal] Refactor config process

tangrammer:
# Description

I'm trying here following:
- Decouple `sdk-config` (the one api receives on sdk/init), from the sdk `system-config` one.
Why? To improve understanding on how config values flow from received, transformed, merged with default configs to finally generate/return a sdk(-system)
pipeline:
`input-config` => `sdk-config` => `system-config` => `sdk-system` 

- also keeping cljc as much of possible to easy env integration ... But this point I'm not really sure if makes sense 🤔


tangrammer:
close in favor of well aligned with main https://github.com/kubelt/kubelt/pull/334

PR #336, opened 2022-04-22
chore(spec-macro): Simplify calling spec conform logic

tangrammer:
the idea is use different validations inside any logic ... (without having several nested if/elses)

Screenshot 2022-04-22 at 12 47 46

### DIFF
--- a/src/main/com/kubelt/lib/config/system.cljc
+++ b/src/main/com/kubelt/lib/config/system.cljc
@@ -28,7 +28,7 @@
 (def default
   {;; These empty defaults should be overridden from the SDK init
    ;; options map.
-   :log/level nil
+   :log/level {:min-level :info}
    :ipfs/read-addr nil
    :ipfs/read-scheme nil
    :ipfs/write-addr nil
--- a/src/main/com/kubelt/lib/config/system.cljc
+++ b/src/main/com/kubelt/lib/config/system.cljc
@@ -3,6 +3,7 @@
   {:copyright "©2022 Proof Zero Inc." :license "Apache 2.0"}
   (:require
    [integrant.core :as ig]
+   [com.kubelt.lib.config.opts :as config.opts]
    [com.kubelt.lib.vault :as lib.vault]
    [com.kubelt.lib.wallet :as lib.wallet]))
 
@@ -28,7 +29,7 @@
 (def default
   {;; These empty defaults should be overridden from the SDK init
    ;; options map.
-   :log/level {:min-level :info}
+   :log/level {:min-level config.opts/log-level}
    :ipfs/read-addr nil
    :ipfs/read-scheme nil
    :ipfs/write-addr nil
--- a/src/main/com/kubelt/lib/error.cljc
+++ b/src/main/com/kubelt/lib/error.cljc
@@ -28,7 +28,10 @@
 (defn explain
   "Return human-friendly description of why some value doesn't conform to
   the provided schema."
-  [schema value]
+  [schema value & [type-error]]
   (let [explain (-> schema (mc/explain value) me/humanize)]
-    {:com.kubelt/type :kubelt.type/error
-     :error explain}))
+    (merge
+     {:com.kubelt/type (or type-error :kubelt.type/error)}
+     (when (:title (mc/properties schema))
+       {:title (:title (mc/properties schema))})
+     {:error explain})))
--- a/src/main/com/kubelt/lib/error.cljc
+++ b/src/main/com/kubelt/lib/error.cljc
@@ -28,10 +28,10 @@
 (defn explain
   "Return human-friendly description of why some value doesn't conform to
   the provided schema."
-  [schema value & [type-error]]
+  [schema value]
   (let [explain (-> schema (mc/explain value) me/humanize)]
     (merge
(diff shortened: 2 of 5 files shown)


src/main/com/kubelt/lib/config/system.cljc
```diff
@@ -28,7 +28,7 @@
 (def default
   {;; These empty defaults should be overridden from the SDK init
    ;; options map.
-   :log/level nil
+   :log/level {:min-level :info}
```
crimeminister, line 31: We currently set a default log level (`:warn`) in `lib.config/log-level` that should be extracted and set as this value:

```
        log-level (get options :log/level)
        ;; Update the system configuration map before initializing the
        ;; system.
        system (-> system
                   (assoc :log/level log-level)
```

 The intent is to define all the defaults together in one place, and to treat all the values extracted from the options map uniformly: extracting them from the options map, setting up the initial values in the system map in the `init` function call as above.

src/main/com/kubelt/sdk/v1.cljs
```diff
@@ -10,7 +10,8 @@
    [com.kubelt.lib.error :as lib.error]
    [com.kubelt.lib.init :as lib.init]
    [com.kubelt.lib.promise :refer [promise promise?]]
-   [com.kubelt.spec.config :as spec.config]))
+   [com.kubelt.spec.config :as spec.config])
+  (:require-macros [com.kubelt.spec :as kspec]))
```
crimeminister, line 14:
```suggestion
  (:require-macros
    [com.kubelt.spec :as kspec]))
```
Minor style thing for consistency with the other `require`s.

src/main/com/kubelt/lib/error.cljc
```diff
@@ -28,7 +28,10 @@
 (defn explain
   "Return human-friendly description of why some value doesn't conform to
   the provided schema."
-  [schema value]
+  [schema value & [type-error]]
```
crimeminister, line 31: I am not sure I understand the intent of allowing the error type to be overridden?

src/main/com/kubelt/spec/config.cljc
```diff
@@ -51,6 +52,17 @@
    [:p2p/multiaddr {:optional true} multiaddr]
    [:p2p/scheme {:optional true} spec.http/scheme]])
 
+(def sdk-config
```
crimeminister, line 55: I am not sure I understand why we want to split the config schema to make a version that supports all optional key/value pairs, and another that makes the values required?

crimeminister (approved):

src/main/com/kubelt/lib/error.cljc
tangrammer: so my intent was to be more specific about error-type if provided, so something like `:kubelt.type/spec-error`
but not sure if `:com.kubelt/type` is opened to that

src/main/com/kubelt/spec/config.cljc
tangrammer: IMO having 2 schemas is powerful! :) 
a couple of reasons:
- we are ensuring defaults are properly merged when they are needed
having every function args "speced" (exact shape of data that internal impl expects) improve not also validation but understanding of the api (internal and public ones) 🤓 
- we can mock/generate sdk values but also system ones

src/main/com/kubelt/spec.clj
```diff
@@ -0,0 +1,9 @@
+(ns com.kubelt.spec
+  (:require
+   [malli.core :as m]
+   [com.kubelt.lib.error :as lib.error]))
+
+(defmacro conform [spec data & body]
+  `(if (m/validate ~spec ~data)
+     (do ~@body)
+     (lib.error/explain ~spec ~data :kubelt.type/spec-error)))
```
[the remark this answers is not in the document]
tangrammer: your comment makes fully sense, I'll take a look to that lib and I'll be back on this topic

src/main/com/kubelt/lib/config/system.cljc
tangrammer: solved by https://github.com/kubelt/kubelt/pull/336/commits/241de288aec23bd35c7923cab6c07973f9f1fc64

src/main/com/kubelt/lib/error.cljc
tangrammer: I'll revert this change until we can discuss and develop better the proposal of error taxonomies and/or using anomalies lib

tangrammer: https://github.com/kubelt/kubelt/pull/336/commits/831e3cdd34970ac42a6287e9741b973f3ea58a1a

src/main/com/kubelt/spec.clj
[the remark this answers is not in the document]
tangrammer: reverting the change until further discussion on this interesting design topic :) 
https://github.com/kubelt/kubelt/pull/336/commits/831e3cdd34970ac42a6287e9741b973f3ea58a1a

tangrammer:
merging now thus the changes will target chore/refactor-sdk-config branch,
thanks for the review!!

PR #339, opened 2022-04-22
feat(sdk): enable injection of platform storage dependency

crimeminister:
# Description

In order to mask platform differences, inject a storage capability into the SDK via the options map. This map should conform to a simple protocol that enables SDK state to be serialized into the environment, e.g for node, the filesystem; for browser, something like local storage. There should also be support for restoring this state into an SDK instance.

- [x] adds an `:app/name` value to the options map passed to `(sdk/init)`
  - this means that the `0-arity` call to `sdk/init` will now result in an error ("missing required key app/name") 
  - this name is used to determine where stored state is kept, i.e. as part of the path to the serialized state file in node, or as part of the key(s) in browser storage where the state is written
- `ddt`
  - [x] add `ddt storage` command group
    - [x] add `ddt storage create` command
      - initializes storage
    - [x] add `ddt storage store` command
      - `$ ddt storage store --wallet default`
      - writes some SDK state into a JSON file
    - [x] add `ddt storage restore` command
      - `$ ddt storage restore`
      - loads some SDK state into an instance
- `sdk`
  - [x] discard "no-op" wallet placeholder
  - [x] introduce `:app/name` required key in `sdk/init` options 
  - `browser`
    - [x] store state using WebStorage `localStorage`  
  - `node` 
    - [x] store state in transit-encoded file

## Type of change


tangrammer:
rebased with fixed conflicts

tangrammer:
@crimeminister 
any lib suggestions for each env implementation?
after a quick search i found following:
- html5 web storage => https://github.com/alandipert/storage-atom
- node => https://github.com/pkpkpk/cljs-node-io
- jvm => db choice?

crimeminister:
I think we should get feedback from @drew-patel about how he'd prefer to store SDK state in the browser and select an option here accordingly, i.e. should we be aiming for IndexedDB, local storage, etc. and should the stored data look like native JSON, or is serialized edn workable, etc.? It looks as though recent versions of `storage-atom` use `transit` as the storage format, which is interesting assuming that it works for Drew.

Another question is about the usage of an atom as with `storage-atom`; as things stand now we have a plain map to represent our system state, and that'd be easy enough to convert to JSON for storage. Is there an advantage to introducing an intermediate atom? That said, this library looks to offer a nice convenient wrapper around local storage, and there's nothing to say we can't have multiple implementations of the `Storage` protocol to account for different requirements even just in-browser, let alone across platforms. As long as we get the protocol right I don't think the implementation should be very heavy; trying a couple of options if we want to should be low cost.

As far as the node option goes (`cljs-node-is`), given the relative simplicity of the data we're storing I am not sure I see too much utility in introducing a dependency to mask the node file APIs to make them more like the JVM, given also that Node is more of a primary concern for us (right now) than JVM. With that in mind, I'd like to defer implementing this protocol for CLJ until the other two platforms are done, and we can discuss some of the many interesting options available (I'd enjoy using XTDB if that made sense, but for such simple data it'd likely be overkill).

drew-patel:
As for the code, the only comment I have is if it's possible to remove most of the requires until we start to use a bunch of them.

drew-patel:
> @crimeminister any lib suggestions for each env implementation? after a quick search i found following:
> 
> * html5 web storage => https://github.com/alandipert/storage-atom
> * node => https://github.com/pkpkpk/cljs-node-io
> * jvm => db choice?

I was thinking if there was an option which we could consider that had APIs for most of not all our targets.

@crimeminister's idea of XTDB is cool. I have worked with Datomic so that would be straightforward to use. I think thing I'm concerned about it how much are we looking to store here?

Do we need to generate a graph of cores for the user locally so we can traverse them and add information to one or many? If so, something like XTDB would be nice.

crimeminister:
@drew-patel We recently added support for extracting an "options" map from an SDK instance to allow it to be stored, and later on passed `sdk/init` to recreate an instance of the SDK. That simple data is all that we're thinking about storing here; it's more about improving the user experience by providing an out-of-box dependency they can inject into the SDK that puts the data in the right place, in the right format for the dev's platform without them needing to do anything beyond an SDK call, e.g. `sdk.v1.store(kbt)`, `sdk.v1.restore(data)`. Right now the user has to call `sdk.v1.options` to get the data, store it themselves, retrieve it themselves when they need it, and then create a new SDK instance in the new context.

If we eventually need to support more complex data storage, querying, and retrieval I imagine we may want a separate implementation / protocol to support that. Perhaps this protocol we're discussing here should be `ConfigStorage` or something like that to communicate the smaller scope that is intended?

drew-patel:
@crimeminister Gotcha! I think the naming did throw me off. `ConfigStorage` is definitely more appropriate.

On the browser, I'm happy to store this in local storage as JSON. Don't need to make things more complicated than that. Thanks for the clarification!

crimeminister:
Quads into facts should be feasible, under the hood XTDB seems to bear some resemblance to a quad store, converting data documents into a collection of facts to be asserted. The connection to IPLD doesn't seem as clear cut, will need to think about what that conversion would look like.

src/main/com/kubelt/spec/storage.cljc
```diff
@@ -0,0 +1,24 @@
...
+   [:cat :any] [:or :string :map]])
+
+(def restore-fn
+  [:=> {:description "A storage function that returns an SDK instance
+using state loaded from platform storage, or an error map if an error
+occurs."}
+   ;; TODO returns a system map (use more precise spec once available)
+   [:cat :any] :map])
+
+(def storage
```
crimeminister, line 19: @tangrammer This was the simple approach I had in mind for wrapping platform configuration storage. This is essentially the same pattern we're using to wrap wallets, i.e. injecting a map with functions that perform the necessary work in a platform-specific way.

src/main/com/kubelt/lib/storage/node.cljs
```diff
@@ -0,0 +1,36 @@
...
+   [malli.core :as malli]
+   [malli.error :as me]
+   [taoensso.timbre :as log])
+  (:require
+   [com.kubelt.lib.error :as lib.error]
+   [com.kubelt.lib.json :as lib.json]
+   [com.kubelt.lib.promise :refer [promise]]
+   [com.kubelt.spec.storage :as spec.storage]))
+
+(defn- store-fn
```
crimeminister, line 16: This state storage function (in node) would be responsible for serializing the state into a file in a well-known location in the filesystem (there's a utility namespace for getting these paths in `com.kubelt.lib.path`). The inverse operation, deserializing the state from a file, would happen in `restore-fn`. This relies more on conditional compilation for cross-platform support rather than protocols as we don't need the dynamic dispatch of protocols given that we have separate builds.

src/main/com/kubelt/lib/config/system.cljc
```diff
@@ -2,26 +2,30 @@
...
 ;; -----------------------------------------------------------------------------
 
-(defn config [system-config options]
+(defn config
+  [system-config options]
   "Return an integrant system configuration map that combines a default
   configuration and a user-provided configuration options map."
   (->> [[:log/level :log/level]
-        [:ipfs.read/multiaddr :ipfs.read/multiaddr]
+        [:app/name :app/name]
```
crimeminister, line 18: @tangrammer I added this as a required parameter to `(init)`. This app name is used to namespace stored data (in case a developer builds multiple apps on the SDK).

src/main/com/kubelt/lib/init.cljc
```diff
@@ -215,13 +222,23 @@
 
 (defmethod ig/init-key :crypto/wallet [_ wallet]
   (log/debug {:log/msg "init wallet"})
-  (if-not (lib.wallet/valid? wallet)
-    (throw (ex-info "invalid wallet" wallet))
+  (if-not (lib.wallet/wallet? wallet)
+    (lib.wallet/create)
```
crimeminister, line 226: @tangrammer Before we were injecting a default "no-op" wallet that had the right shape to pass validation, but didn't perform crypto operations. With this change we either use an injected wallet or we construct a real wallet instance instead, without creating a temporary placeholder.

src/main/com/kubelt/lib/init.cljc
```diff
@@ -215,13 +222,23 @@
...
 
 (defmethod ig/halt-key! :crypto/wallet [_ wallet]
   (log/debug {:log/msg "halt wallet"}))
 
+;; config/storage
+;; -----------------------------------------------------------------------------
+
+(defmethod ig/init-key :config/storage [_ {app-name :app/name :as storage}]
+  (log/debug {:log/msg "init storage" :app/name app-name})
+  (lib.storage/create app-name))
```
crimeminister, line 237: @tangrammer Right now we create a platform-specific storage capability; in the future we can accept a developer-injected storage capability as well.

src/main/com/kubelt/lib/storage/browser.cljs
```diff
@@ -0,0 +1,16 @@
...
+  (:require
+   [malli.core :as malli])
+  (:require
+   [com.kubelt.lib.error :as lib.error]
+   [com.kubelt.lib.json :as lib.json]
+   [com.kubelt.lib.promise :refer [promise]]
+   [com.kubelt.proto.storage :as proto.storage]
+   [com.kubelt.spec.storage :as spec.storage]))
+
+(defrecord Storage []
```
crimeminister, line 13: @tangrammer In the node implementation so far I have done without using a protocol. Dynamic dispatch doesn't seem necessary as long as we are using a reader conditional to build multiple versions of the SDK for different platforms. Is there some additional advantage to using a protocol for this?

src/main/com/kubelt/spec/config.cljc
```diff
@@ -61,10 +70,16 @@
 ;; should have an SDK configuration options map that has every value
 ;; provided.
 (def sdk-config
-  (into [:map {:closed true
-               :title ::sdk-config}]
-        (map #(assoc-in % [1 :optional] false)
-             (m/-children (m/schema optional-sdk-config nil)))))
+  (let [;; These keys remain optional; if not provided by the user, we
+        ;; initialize their values to platform defaults in lib.init.
+        excluded-keys #{:config/storage :crypto/wallet}]
```
crimeminister, line 75: @tangrammer This change avoids marking the keys `#{:config/storage :crypto/wallet}` as required when validating the merged configuration. The idea here is that, if the user doesn't inject a value for these keys, then the appropriate place to create the value is during the execution of integrant `init`, without having to supply an intermediate placeholder to pass validation and then throwing it away.

src/main/com/kubelt/spec/storage.cljc
```diff
@@ -0,0 +1,26 @@
+(ns com.kubelt.spec.storage
```
crimeminister, line 1: @tangrammer Instead of a protocol, so far I am getting away with using a map of the shape described by this schema.

src/main/com/kubelt/lib/storage/browser.cljs
tangrammer: don't think so

src/main/com/kubelt/spec/config.cljc
tangrammer: yep, i got the logic of this impl, anyway this is one of the hard to understand parts of the codebase, meaning that to config we filter in one place, use defaults in a other place and init the "final" values in other place 

start to thinking that we need to define different structures/defs depending of the meaning that these structures represent ("Thinking aloud TM" 😄  )

eg: optional-config > config > unauthenticated-system > authenticated-system

crimeminister:
@alfl If we decide on some additional requirements for this work, e.g. encrypting stored data, lmk svp.

alfl (dismissed):

alfl (approved):

PR #392, opened 2022-05-12
Fix warning: ns-var-clash

tangrammer:
Fix following warning 

```
------ WARNING - :ns-var-clash -------------------------------------------------
 File: /Users/tangrammer/git/kubelt/kubelt/src/main/com/kubelt/spec/openrpc.cljc:19:1
 Namespace com.kubelt.spec.openrpc.schema clashes with var com.kubelt.spec.openrpc/schema
--------------------------------------------------------------------------------
````

the problem also prevents code evaluation in dev local env, so fixing it enable normal dev REPL flow


crimeminister:
This may be a problem of variable capture in the `conform*` macro, will take a quick look to see if it's something that can be fixed there.

PR #376, opened 2022-05-04
feat(sdk): support for JSON-RPC client

crimeminister:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

Closes # (issue)

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires or is a documentation update
- [ ] Security Fix

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation website
- [ ] I have made corresponding changes to the literal docs
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #393, opened 2022-05-12
Feat/rpc client ddt tasks

tangrammer:
Relates #376 to be merged in https://github.com/kubelt/kubelt/tree/feat/rpc-client
### `ddt rpc `
- [x] add `call` command for calling provider RPC endpoints
    
```
node ddt.js rpc call :kb:ping  -w default  -x value_1=1 -x value_2=2
Enter your password:  ******

com.kubelt.ddt.js default
:method kb_ping :params {:value_1 1, :value_2 2}
call result: pong
```

- [x] add `list` command for listing available RPC endpoints

```
> node ddt.js rpc list -w default

Enter your password:  ******
#{[:kb :auth] [:kb :ping] [:kb :pong] [:kb :auth :verify]}
```

### DIFF
--- a/src/main/com/kubelt/ddt/cmds/rpc/call.cljs
+++ b/src/main/com/kubelt/ddt/cmds/rpc/call.cljs
@@ -65,5 +65,6 @@
                                                                         (println :method rpc-method
                                                                                  :params rpc-params)
                                                                         (println "call result: " r)))))))
-                          (lib.promise/catch (fn [error]
-                                               (println "we got an error" error))))))))))})
+                          (lib.promise/catch (fn [e]
+                                               (println (ex-message e))
+                                               (prn (ex-data e)))))))))))})
--- a/src/main/com/kubelt/ddt/cmds/rpc/call.cljs
+++ b/src/main/com/kubelt/ddt/cmds/rpc/call.cljs
@@ -3,13 +3,14 @@
   {:copyright "ⓒ2022 Proof Zero Inc." :license "Apache 2.0"}
   (:require
    [clojure.string :as cstr]
+   [com.kubelt.ddt.cmds.sdk.core.authenticate :as ddt.auth]
    [com.kubelt.ddt.options :as ddt.options]
    [com.kubelt.ddt.prompt :as ddt.prompt]
    [com.kubelt.ddt.util :as ddt.util]
    [com.kubelt.lib.promise :as lib.promise]
+   [com.kubelt.rpc :as rpc]
    [com.kubelt.sdk.v1.core :as sdk.core]
-   [com.kubelt.ddt.cmds.sdk.core.authenticate :as ddt.auth]
-   [com.kubelt.rpc :as rpc]))
+   [taoensso.timbre :as log]))
 
 (defonce command
   {:command "call <method>"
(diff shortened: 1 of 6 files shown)


src/main/com/kubelt/ddt/cmds/sdk/core/authenticate.cljs
```diff
@@ -13,6 +13,32 @@
    [com.kubelt.sdk.v1 :as sdk]
    [com.kubelt.sdk.v1.core :as sdk.core]))
 
+(defn authenticate [args-map password on-authenticate]
```
tangrammer, line 16: refactor: just extracting logic outside so we can reuse/call it from other places

src/main/com/kubelt/lib/p2p.cljs
```diff
@@ -14,35 +15,55 @@
...
-    (str (name scheme) "://" host ":" port path)))
+        path (cstr/join "" ["/@" core "/jsonrpc"])
+        uri (str (name scheme) "://" host ":" port path)]
+    (log/debug {::connection-uri uri})
+    uri))
+
+(defn- json-rpc-provider [sys core]
+  (JsonRpcProvider. (connection-uri sys core)))
+
+(defn- send
```
tangrammer, line 26: more refactors and so far all json-rpc logic is being concentrated here but feel free to move it to proper ns

src/main/com/kubelt/ddt/cmds/rpc/call.cljs
```diff
@@ -20,10 +24,9 @@
                                 :describe "An RPC parameter (<name>=<value>)"
                                 :requiresArg true
                                 :demandOption "param name and value are required"
-                                :string true
+                                :array true
```
tangrammer, line 27: do we agree on use this type so we can send more than 1 args?
each one should be prefixed ...
eg: `-x value_1=1 -x value_2=2`

crimeminister: I did some experimentation and thought I remembered multiple parameter values being collected even when I had `:string true` set, but I just did a little test and it blew up. Looks like this change is necessary, thanks.

src/main/com/kubelt/ddt/cmds/sdk/core/authenticate.cljs
```diff
@@ -13,6 +13,32 @@
...
+ (let [app-name (get args-map :app-name)
+       wallet-name (get args-map :wallet)]
+   (println app-name wallet-name)
```
crimeminister, line 19: Should we remove this `(println)`?

tangrammer: yep!

src/main/com/kubelt/ddt/cmds/sdk/core/authenticate.cljs
crimeminister: Sounds good!

src/main/com/kubelt/lib/p2p.cljs
crimeminister: After this lands I will look at moving the RPC client into the SDK state map, and we can re-use it that way. That'll give us the option of injecting an RPC client implementation as well, so we can experiment with different options there (at a minimum, choose between `JsonRpcProvider` and our local client).

crimeminister (approved):

PR #394, opened 2022-05-16
chore(refactor) Use promises and fs-promises

tangrammer:
trying to standardise the use of promises a bit more using fs node support

### DIFF
--- a/src/main/com/kubelt/ddt/cmds/wallet/import.cljs
+++ b/src/main/com/kubelt/ddt/cmds/wallet/import.cljs
@@ -34,7 +34,7 @@
                         (let [password (.-password result)
                               result& (lib.wallet/import app-name wallet-name mnemonic password)]
                           (-> result&
-                              (.then (fn [{:keys [wallet/name wallet/path]}]
+                              (.then (fn [{:keys [wallet/name]}]
                                        (let [message (str "wallet '" name "' successfully imported")]
                                          (println message))))
                               (.catch (fn [error]
--- a/src/main/com/kubelt/ddt/cmds/wallet/import.cljs
+++ b/src/main/com/kubelt/ddt/cmds/wallet/import.cljs
@@ -18,12 +18,6 @@
                     app-name (get args-map :app-name)
                     wallet-name (get args-map :name)]
                 ;; TODO check to see if wallet name is valid (using yargs)
-
-                ;; Check to see if named wallet already exists. Wallets
-                ;; can't be overwritten so throw an error if so.
-                (when (lib.wallet/has-wallet? app-name wallet-name)
-                  (let [message (str "wallet '" wallet-name "' already exists")]
-                    (ddt.util/exit-if message)))
                 (ddt.prompt/ask-mnemonic!
                  (fn [err result]
                    (let [mnemonic (.-mnemonic result)]
@@ -32,7 +26,7 @@
                         (when err
                           (ddt.util/exit-if err))
                         (let [password (.-password result)
-                              result& (lib.wallet/import app-name wallet-name mnemonic password)]
+                              result& (lib.wallet/import& app-name wallet-name mnemonic password)]
                           (-> result&
                               (.then (fn [{:keys [wallet/name]}]
                                        (let [message (str "wallet '" name "' successfully imported")]
--- a/src/main/com/kubelt/ddt/cmds/wallet/init.cljs
+++ b/src/main/com/kubelt/ddt/cmds/wallet/init.cljs
@@ -4,7 +4,6 @@
   (:require
    [com.kubelt.ddt.options :as ddt.options]
    [com.kubelt.ddt.prompt :as ddt.prompt]
-   [com.kubelt.ddt.util :as ddt.util]
    [com.kubelt.lib.promise :as lib.promise]
    [com.kubelt.lib.wallet :as lib.wallet]))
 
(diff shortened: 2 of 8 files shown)


src/main/com/kubelt/ddt/cmds/wallet/import.cljs
```diff
@@ -18,12 +18,6 @@
                     app-name (get args-map :app-name)
                     wallet-name (get args-map :name)]
                 ;; TODO check to see if wallet name is valid (using yargs)
-
-                ;; Check to see if named wallet already exists. Wallets
-                ;; can't be overwritten so throw an error if so.
-                (when (lib.wallet/has-wallet? app-name wallet-name)
```
tangrammer, line 24: this check was already performed in lib.wallet side

src/main/com/kubelt/ddt/cmds/wallet/init.cljs
```diff
@@ -21,26 +19,16 @@
                     app-name (get args-map :app-name)
                     ;; TODO check to see if wallet name is valid (using yargs)
                     wallet-name (get args-map :name)]
-                (lib.promise/promise
-                 (fn [resolve reject]
-                   ;; Check to see if named wallet already exists. Wallets
-                   ;; can't be overwritten so throw an error if so.
-                   (if (lib.wallet/has-wallet? app-name wallet-name)
```
tangrammer, line 28: moved this condition to https://github.com/kubelt/kubelt/blob/fix%2Finit-wallet-use-promise/src/main/com/kubelt/lib/wallet/node.cljs#L114

src/main/com/kubelt/ddt/cmds/wallet/import.cljs
crimeminister: Do we need to add an additional `(.catch)` to manage a promise rejected in `lib.wallet/import`?

tangrammer: updated to use lib.promise/catch

src/main/com/kubelt/ddt/cmds/wallet/init.cljs
```diff
@@ -21,26 +19,16 @@
...
-                     (ddt.prompt/confirm-password!
-                      (fn [err result]
-                        (when err
-                          ;;(ddt.util/exit-if err)
-                          (reject err))
-                        (let [password (.-password result)]
+                (ddt.prompt/confirm-password!
+                 (fn [err result]
+                   (when err
+                     (println err))
```
crimeminister, line 25: If an error occurs for password confirmation then the passwords didn't match, so should we exit with an error in that  case, perhaps using the `(exit-if)` utility?

src/main/com/kubelt/ddt/cmds/wallet/init.cljs
```diff
@@ -21,26 +19,16 @@
...
+                   (-> (lib.wallet/init& app-name wallet-name (.-password result))
+                       (lib.promise/then
+                        (fn [wallet]
                           (let [;; TODO return a promise from wallet/init
-                                wallet (lib.wallet/init app-name wallet-name password)
                                 wallet-name (get wallet :wallet/name)
-                                ;; NB: Also available are the mnemonic path, locale.
+                     ;; NB: Also available are the mnemonic path, locale.
```
crimeminister, line 31: Small indentation issue.

src/main/com/kubelt/lib/wallet/node.cljs
```diff
@@ -64,111 +73,120 @@
...
-          false)))
-    false))
+  (-> (has-wallet?& app-name wallet-name)
+      (lib.promise/then
+       (fn [wallet-path]
+         (->
+          (.readFile fs-promises wallet-path)
+          (lib.promise/then #(.fromEncryptedJson Wallet % password))
+          (lib.promise/then (fn [_] true))
+          (lib.promise/catch (fn [e]
```
crimeminister, line 92: Should we avoid catching an error here, leaving it to the caller to supply exception handling behavior? If we mask errors that occur here, will the caller have another path to discovering what went wrong, e.g. supplied an invalid path to a wallet, rather than an invalid password that prevents successful decryption. 🤔

src/main/com/kubelt/ddt/cmds/wallet/init.cljs
tangrammer: done!

src/main/com/kubelt/lib/wallet/node.cljs
```diff
@@ -16,25 +16,39 @@
...
+
 (defn- wallet-dir
   "Return the wallet directory path as a string for an application."
   [app-name]
   (let [config-path (lib.path/data app-name)
         wallet-path (.join path config-path "wallets")]
     wallet-path))
 
-(defn- ensure-wallet-dir
+(defn- ensure-wallet-dir&
```
tangrammer, line 35: only `ensure_` prefixed funs reject

src/main/com/kubelt/lib/wallet/node.cljs
tangrammer: after a few more work on how promises work and how we did, I think code is better now

crimeminister:
@tangrammer Looks like we need to fix up an issue in the build pipeline so the tests go green.

PR #404, opened 2022-05-19
RPC unit tests and ddt rpc adaptations

tangrammer:
adding rpc-client unit tests 
and adapt ddt rpc commands to use rpc-client

to be merged on [feat/rpc-client](https://github.com/kubelt/kubelt/tree/feat/rpc-client)


fix/openrpc/be.edn
```diff
@@ -0,0 +1,46 @@
+{:openrpc "1.2.6",
```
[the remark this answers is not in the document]
tangrammer: nope .. doing now ...

src/main/com/kubelt/ddt/cmds/rpc/call.cljs
```diff
@@ -2,17 +2,25 @@
   "Make an RPC call."
   {:copyright "ⓒ2022 Proof Zero Inc." :license "Apache 2.0"}
   (:require
-   [clojure.string :as cstr])
-  (:require
-   [taoensso.timbre :as log])
-  (:require
+   [clojure.string :as cstr]
```
crimeminister, line 5: Not terribly important, but as a matter of style I usually group requires from stdlib separately from external dependpencies, separately from internal namespaces.

src/test/com/kubelt/rpc/schema_test.cljs
```diff
@@ -0,0 +1,86 @@
+(ns com.kubelt.rpc.schema-test
```
crimeminister, line 1: Later I'd like to change this to `.cljc` and make sure these tests work with `clj` as well. Probably makes more sense to bother with after we change our promise implementation though.

crimeminister (approved):

PR #407, opened 2022-05-19
feat(sdk): Use project.clj to build JAR for CLJ/CLJS apps

drew-patel:
# Description

In order to use the SDK as a dependency via `shadow-cljs`, it is simpler to call `lein install` and build against it.

For CLJS browser apps, they would need to all the following in their compiler options map to have the SDK work as intended:

```
{:compiler-options
  {:reader-features #{:browser}
   ...}}
```

Eventually this JAR can be used as an artifact to build and test other products that are created.

## Type of change

# How Has This Been Tested?

- [x] Built JAR locally and used in `three-id` repo 

# Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -166,10 +166,7 @@ dist
 # Leiningen
 /.lein-*
 /.nrepl-port
-# We use the leiningen project file only to aid in packing up a CLJS
-# library (as a JAR). The project.clj is generated from shadow-cljs.edn
-# so no need to check it in.
-project.clj
+pom.xml
 
 # clj-kondo cache
 /.clj-kondo/.cache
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -166,10 +166,6 @@ dist
 # Leiningen
 /.lein-*
 /.nrepl-port
-# We use the leiningen project file only to aid in packing up a CLJS
-# library (as a JAR). The project.clj is generated from shadow-cljs.edn
-# so no need to check it in.
-project.clj
```
tangrammer, line 169: should we add here pom.xml ?
I saw that it's generated as a result of lein install

drew-patel: Yes. Good catch. Will do!

PR #408, opened 2022-05-20
Refactor lib.error/conform* macro

tangrammer:
changing signature and reusing logic  

trying to be more specific about specs/guards and more flexible about the body

from
`(defmacro conform* [& args]) ` to `(defmacro conform* [guards & body])`

to be merged on [feat/rpc-client](https://github.com/kubelt/kubelt/tree/feat/rpc-client)


src/main/com/kubelt/lib/error.cljc
```diff
@@ -53,39 +57,27 @@
...
-        body
-        (let [[spec data] (first guards)]
-          (recur (rest guards) (validate-if spec data body)))))))
-
-(comment
-  (require '[com.kubelt.spec.rpc :as spec.rpc])
-  (conform* [spec.rpc/path [:eth :sync]] [spec.rpc/path [:web-3 :sha-3]] :foo)
-  (conform* [spec.rpc/path [:eth :sync]] [spec.rpc/path [:web-3 :sha-3]] (+ 1 2)))
+(defmacro conform*
+  [guards & body]
```
crimeminister, line 73:
```suggestion
  "Check that pairs of specs and data are all valid, and if all data
  matches the provided specs then execute the provided body. Otherwise,
  an error is returned that describes the errors found in the first data
  that failed validation."
  [guards & body]
```

crimeminister, line 73: I'd prefer to keep the parameter list as `& args`, and so some work inside the macro to build the list of guards, e.g. 
```clojure
(let [guards# (butlast args)
      body (last args)]
       ..)
```
Stylistically I just think it's a bit nicer to avoid the extra layer of nesting of the `[spec data]` pairs. Afterwards we can add some error checking to validate that the guards and the body have the expected shape, and throw an exception if the macro was used incorrectly.

crimeminister:
@tangrammer I sent PR #411 that does (I think) what is intended by this PR. Let me know if you think further refinement is necessary or would be useful, svp.

PR #412, opened 2022-05-20
feat(sdk): improve SDK schema expansion

crimeminister:
# Description

When loading a schema for use by the client, references are now expanded inline to make later processing easier.

Other changes:
- [x] initial validation; checking for missing refs
- [x] `prepare` accepts RPC params as either vector or map
  - when vector is supplied, parameters are matched positionally
  - when map is supplied, keys are parameter name and map to corresponding value
- [x] further processing of schema method definition
- [x] work on request body generation
- [x] small style tweaks to specs

## Type of change

### DIFF
--- a/src/dev/.gitignore
+++ b/src/dev/.gitignore
@@ -1,0 +1,3 @@
+# src/dev/.gitignore
+
+robert/
--- a/src/dev/user.cljc
+++ b/src/dev/user.cljc
@@ -1,0 +1,9 @@
+(ns user
+  "Shared development utilities."
+  {:copyright "©2022 Proof Zero Inc." :license "Apache 2.0"})
+
+;; development
+;; -----------------------------------------------------------------------------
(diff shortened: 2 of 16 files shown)


drew-patel (dismissed):
:shipit:

tangrammer:
not sure that we should add (git-track) dev/dev-user(s) namespaces 🤔

drew-patel:
> I have learnt from other's dev namespaces in the past, but it can add some churn so I don't mind taking it out.

Feel the same way about learning from other people's dev tooling/config. Maybe we can find another way to share these.

tangrammer:
> > I have learnt from other's dev namespaces in the past, but it can add some churn so I don't mind taking it out.
> 
> Feel the same way about learning from other people's dev tooling/config. Maybe we can find another way to share these.

tangrammer (approved):
LGTM!

PR #417, opened 2022-05-25
feat(profile): spec

### LINKED ISSUE #434: feat(profile): spec
# Why

> It's a good idea to have a representation of the data models in the SDK

# How

- Add profile spec
- Add profile test
- Update profile & config specs with optional elements
- Add minimal test cases for profile & config

Cosmin-Parvulescu:
# Description

> It's a good idea to have a representation of the data models in the SDK

- Added profile spec
- Added profile test
- Updated profile & config specs with optional elements
- Added minimal test cases for profile & config

Closes #434 

## Type of change

# How Has This Been Tested?

- [x] (t/test-all)
- [x] bb test:sdk:cljs:develop
- [x] CI

# Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works


tangrammer (approved):

crimeminister:
@Cosmin-Parvulescu can we delete this branch?

Cosmin-Parvulescu:
> @Cosmin-Parvulescu can we delete this branch?

Yeah, I closed it because on Friday-demo we decided this would not be part of the SDK. Might be worth to keep the branch associated to it, but the PR can def. be deleted.

crimeminister:
Having this schema around _would_ be useful, for educational purposes at least (while the backend service lives outside of the monorepo). I guess the main cost to having it here would be duplication and the need to keep it in sync with upstream changes?

Cosmin-Parvulescu:
> Having this schema around _would_ be useful, for educational purposes at least (while the backend service lives outside of the monorepo). 

This is just thoughtfood but I think there will be a set of types that are going to be fundamental to the backend service, which would be nice to have in the SDK for education et. al., but I'm not sure profile is part of them. I think we can safely skip it for now, and indeed, even remove the branch. :)

> I guess the main cost to having it here would be duplication and the need to keep it in sync with upstream changes?

I think that's a big cost to pay right now if we're going to be using the profile functionality through the SDK ```send``` method, and not something specific. Perhaps when we feel the need to add convenience methods (such as for auth), maybe then it's time to bring in the typings to the SDK, and perhaps even generate typings for clients (whether front or back ends).

PR #418, opened 2022-05-25
Fix shadow-cljs warnings

tangrammer:
# Description

Relates and closes https://github.com/kubelt/kubelt/pull/392

Caused by cljs compilation bug https://github.com/thheller/shadow-cljs/issues/874... so renaming a def https://github.com/kubelt/kubelt/compare/fix-shadow-warnings?expand=1#diff-b862c75b84e447a6ed4ced02513a37f7a0884c1151b0f1a31de8c00e9e77d3b1R19

### DIFF
--- a/src/main/com/kubelt/rpc/request.cljc
+++ b/src/main/com/kubelt/rpc/request.cljc
@@ -84,7 +84,7 @@
                         (if (lib.error/error? result)
                           [param-kw result]
                           pair))
-                      [param-kw (lib.error/error "no such parameter" :parameter param-kw)]))
+                      [param-kw (lib.error/error ["no such parameter" :parameter param-kw])]))
                   params))))
 
 ;; from-params-vec
--- a/src/main/com/kubelt/rpc/schema/zip.cljc
+++ b/src/main/com/kubelt/rpc/schema/zip.cljc
@@ -35,7 +35,7 @@
 
 (defn- map-entry
   [k v]
-  (clojure.lang.MapEntry/create k v))
+  #?(:clj (clojure.lang.MapEntry/create k v)))
 
 ;; branch?
 ;; -----------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


src/main/com/kubelt/spec/openrpc.cljc
```diff
@@ -16,7 +16,7 @@
 ;; Root
 ;; -----------------------------------------------------------------------------
 
-(def schema
+(def schemata
```
crimeminister, line 19: In the docs they sometimes refer to this as the "root object". Okay if we use `root` as this var name to keep consistent?

src/main/com/kubelt/rpc/schema/zip.cljc
```diff
@@ -35,7 +35,7 @@
 
 (defn- map-entry
   [k v]
-  (clojure.lang.MapEntry/create k v))
+  #?(:clj (clojure.lang.MapEntry/create k v)))
```
crimeminister, line 38: Hrmm, I assumed there was a `clojure.lang.MapEntry` in CLJS as well. Will have to look into what we should be using on that platform instead.

src/main/com/kubelt/rpc/request.cljc
```diff
@@ -84,7 +84,7 @@
                         (if (lib.error/error? result)
                           [param-kw result]
                           pair))
-                      [param-kw (lib.error/error "no such parameter" :parameter param-kw)]))
+                      [param-kw (lib.error/error ["no such parameter" :parameter param-kw])]))
```
crimeminister, line 87: We should probably add another arity to `lib.error/error` to make it consistent with `ex-info`, e.g.
```clojure
(error msg)
(error msg map)
(error msg map cause)
```

tangrammer: adapted function

src/main/com/kubelt/rpc/schema/zip.cljc
crimeminister: https://cljs.github.io/api/cljs.core/MapEntry

crimeminister (approved):

PR #419, opened 2022-05-25
chore(gpg): add public signing key

crimeminister:
# Description

Adds public release signing key to repo.


alfl (approved):
Checked first and last byte substrings against public key in 1pass. LGTM.

PR #421, opened 2022-05-25
feat(rpc): memoize $ref lookup fn

crimeminister:
# Description

Memoizes the `rpc.schema.util/lookup` fn that takes a `$ref` and an edn schema and (possibly recursively) looks up the referenced definition.


tangrammer (commented):
perhaps better to use the decorator approach https://stackoverflow.com/questions/53644511/what-is-the-recommended-way-to-use-memoize

then we only need to add in the same ns  `(alter-var-root #'lookup memoize)` 

perhaps better when  doing local dev/test without memoization

crimeminister:
@tangrammer Nice tip, changing this.

tangrammer (approved):

PR #432, opened 2022-05-26
feat(rpc): support OpenRPC schema discovery

crimeminister:
# Description

OpenRPC supports a method for discovering the schema that defines a compliant API. Our backend is supporting that mechanism, enabling us to dynamically discover what calls are available. This adds initial support for this discovery mechanism.

- add `com.kubelt.lib.uri` namespace for cross-platform URI utilities
  - contains capabilities required to work with OpenRPC Server configuration
  - [x] `parse` URI string into our map format
  - [ ] tests for `parse` fn
  - [ ] `unparse` our map format into URI string
  - [ ] tests for `unparse` fn
  - [ ] `expand` a [URL template](https://datatracker.ietf.org/doc/html/rfc6570) into a string using supplied parameters
  - [ ] tests for `expand` fn
- add `com.kubelt.rpc.server` namespace for working with OpenRPC `Server`

## Type of change

### DIFF
--- a/src/main/com/kubelt/lib/uri.cljc
+++ b/src/main/com/kubelt/lib/uri.cljc
@@ -119,8 +119,8 @@
 
 (defn expand
   "Given a URL template and a map of parameter values, return the expanded
-  URI."
-  [tpl params]
-  {:pre [(string? tpl) (map? params)]}
-  :fixme
-  )
+  URI. If no variables are supplied, the template is returned unchanged."
+  [tpl variables]
+  {:pre [(string? tpl) ((some-fn [nil? map?]) variables)]}
+  ;; TODO actually expand this template
+  tpl)
--- a/src/main/com/kubelt/lib/uri.cljc
+++ b/src/main/com/kubelt/lib/uri.cljc
@@ -1,0 +1,126 @@
+(ns com.kubelt.lib.uri
+  "URI-related utilities."
+  {:copyright "©2022 Proof Zero Inc." :license "Apache 2.0"}
+  (:require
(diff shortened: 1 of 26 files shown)


src/main/com/kubelt/lib/uri.cljc
```diff
@@ -0,0 +1,126 @@
...
+  ""
+  [m]
+  #_(lib.error/conform*
+     [])
+  :fixme
+  )
+
+;; expand
+;; -----------------------------------------------------------------------------
+;; TODO convert a URL template string and associated parameter values
```
crimeminister, line 117: @tangrammer Would you mind taking a look for an existing cross-platform implementation of [URI Templates](https://datatracker.ietf.org/doc/html/rfc6570) we can use from here? Alternately, would you care to explore an internal implementation? For the moment I don't think we need anything more than basic interpolation, but we should keep the door open for a more full compliant implementation down the road.

tangrammer:
rebased

src/main/com/kubelt/lib/uri.cljc
tangrammer: so i found this clj lib https://github.com/mwkuster/uritemplate-clj that seems realistic following the test suite
and about exploration an internal impl ...  I tried to move uritemplate-clj to cljs too, here my WIP => https://github.com/mwkuster/uritemplate-clj/compare/master...tangrammer:master ... my attempt is so far removing clj dep to clj(s) dep (using https://github.com/lambdaisland/uri instead of https://github.com/ring-clojure/ring-codec ) and so far tests are passing

PR #436, opened 2022-05-27
Fix clj compilation

tangrammer:


crimeminister:
As an experiment, I changed our `require` to:
```
(:require
  [com.kubelt.lib.error :as lib.error :refer [conform*] :refer-macros [conform*])
```
with the hope that Clojure would ignore the `:refer-macros` phrase, and that in ClojureScript the `:refer-macros` binding would override the `:refer`. Seems to work in Clojure, but needs more testing.

tangrammer:
tested in cljs your proposal and working fine so adding it to this PR

crimeminister (approved):

PR #438, opened 2022-05-30
chore(spec): add OpenRPC specs

crimeminister:
# Description

Introduces some of the OpenRPC-related schemas from the in-flight branch in case they're useful sooner for other work. Also fixes some lint issues in the perennial quest to become lint-clean.

## Type of change


crimeminister:
I hadn't realized that there would be tests impacted by these changes. Will land the schema changes alongside related code changes.

PR #449, opened 2022-06-01
Use oort-devnet BE in test-commons

tangrammer:


crimeminister:
@szkl Could you let us know when the time is right for this change?

szkl (approved):

PR #451, opened 2022-06-02
Add storage/browser impl

tangrammer:
# Description

Add storage/browser impl

### DIFF
--- a/dapp/src/dapp/core.cljs
+++ b/dapp/src/dapp/core.cljs
@@ -24,7 +24,7 @@
 (re-frame/reg-event-db
  ::initialize-db
  (fn [_db _]
-   (.then (sdk.v1/init)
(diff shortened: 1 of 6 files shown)


tangrammer:
@crimeminister  do we need localStore or sessionStore? 
so far added a dynamic binding to have flexibility but no reason to keep it once we decided the storage-type

crimeminister:
> @crimeminister do we need localStore or sessionStore? so far added a dynamic binding to have flexibility but no reason to keep it once we decided the storage-type

From MDN:
```
The two mechanisms within Web Storage are as follows:

    - sessionStorage maintains a separate storage area for each given origin that's available for the duration of the page session (as long as the browser is open, including page reloads and restores).
    - localStorage does the same thing, but persists even when the browser is closed and reopened.
```
@drew-patel @alfl does one of these options make more sense given the usage patterns you envision in the app?

PR #468, opened 2022-06-08
chore(sdk): Update groupId for Maven to publish to Clojars

drew-patel:
# Description

Clojars only supports groupId's with specific formats. In our case, we have `com.kubelt`.

See: https://github.com/clojars/administration/issues/260

## Type of change

# How Has This Been Tested?

- [x] Pushed successfully to Clojars - https://clojars.org/com.kubelt/sdk

# Checklist:

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(defproject kubelt/sdk "0.0.1"
+(defproject com.kubelt/sdk "0.0.2"
   :description "Kubelt SDK"
   :url "https://kubelt.com/"
 


crimeminister:
@drew-patel This may be an opportune moment to revisit our `bb` tasks that generate a `project.clj` from the `deps.edn` file, e.g. in `packages/sdk-clj`:

```shell
$ bb run generate:clj:project
```
The intent is that we avoid managing dependencies in multiple overlapping places; all deps live in `deps.edn` and `shadow-cljs.edn` and we generate the `project.clj` as-needed during the publishing process, since we're only using that file to manage some of the vagaries of publishing artifacts to public package repositories.

I'll extract the task code that collects dependencies and move into a library, and do a couple of other bits of cleanup. After that, would you be okay with taking this approach of generating the `project.clj` file on-demand?

crimeminister (approved):
LGTM

drew-patel:
@crimeminister Yes. I remember you showed me the `project.clj` generation. Would you like me to get rid of this file now then? 

I can use the generator to generate and then publish the SDK from now on. I just needed "something" on Clojars to get our deployments going

crimeminister:
> @crimeminister Yes. I remember you showed me the `project.clj` generation. Would you like me to get rid of this file now then?

Let's merge this and then collab on a branch that uses `bb` to do the needful. We can compare the generated result against this `project.clj` and only remove it when we've achieved parity. Work for you?

PR #469, opened 2022-06-08
feat(sdk): clean up generation of project.clj

crimeminister:
# Description

Before switching to programmatic generation of `project.clj` in our publishing pipeline, do a bit of clean up and refactoring. The eventual aim is to rely on a small set of files for platform-specific dependencies, ideally without any overlap between them that might allow discrepancies to sneak in:
- `package.json` for JavaScript
- `shadow-cljs.edn` for Clojure/Script

The top-level `deps.edn` file should only be used for tooling-related concerns, e.g. linting, test frameworks, etc. and not for dependency information.

## Changes

- [x] make `bb` version `0.8.0` the minimum required version
- [x] extract task code for generating `project.clj` into shared `bb` library
  - adds `bb/` ns `com.kubelt.lein` with function `make-project`
- [x] extract task code for generating `deps.edn` into shared `bb` library
  - adds `bb/` ns `com.kubelt.deps` with function `make-deps` 
- [x] add `bb/` ns `com.kubelt.version`
  - defines versions of Clojure/Script marked as "provided" in `project.clj`
- [x] add `bb/` ns `com.kubelt.common`
  - defines some constant values 
- [x] updates `bb.edn` files to use library code for some tasks
  - `packages/sdk-clj/bb.edn`
  - `packages/sdk-cljs/bb.edn`  
- [x] add `clojars` publishing configuration to `project.clj`
- [x] specify name of release signing key in `project.clj`
- [x] remove top-level `project.clj` file
- [x] add `project.clj` to `.gitignore` in the project root

## Type of change

- [x] Clean-up and refactoring

## Notes

- the bb library code is located under the top level directory `bb/`


crimeminister:
@drew-patel Do you have any preferences re supporting `SNAPSHOT` releases right now?

bb/com/kubelt/lein.clj
```diff
@@ -0,0 +1,40 @@
...
+
+
+;; A Leiningen project map to serve as the base on which we add
+;; more data, e.g. dependencies.
+(def ^:private base-map
+  {:name :com.kubelt/sdk
+   :description "Kubelt SDK"
+   :url "https://kubelt.com/"
+   :source-paths ["src/main"]
+   :repositories
```
drew-patel, line 19: This is a potential addition but when publishing to either Clojars or Github, we get asked for a `:license` key. Thoughts on what you'd like to include there?

drew-patel (approved):
Couple more comments to wrap up the discussion:

- I'm not opposed to the `SNAPSHOT` way of releasing. We should probably have a discussion on how often we'll choose to release the SDK and coordinating with other products. I think for now, we don't need it since we'll be the only consumers

- `shadow-cljs` has a way to use `deps.edn` for dependencies. How do you feel about that? https://shadow-cljs.github.io/docs/UsersGuide.html#deps-edn (I'm not suggesting we do that in this PR)

bb/com/kubelt/lein.clj
crimeminister: I'll add a `:license` field with a value that matches what's in our file headers, thanks.

crimeminister: Updated!

PR #473, opened 2022-06-09
fix(sdk): improve JS handling of SDK init options

crimeminister:
# Description

Adds some JavaScript tests of SDK initialization where we supply various subsets of configuration options, e.g. connection details for p2p, IPFS read and write nodes, etc. Fixes an issue in the conversion of the JS configuration object into a CLJ(S) configuration map so that the results conform to our schema. Specifically, some configuration values provided in JavaScript (e.g. `{"ipfs.read/scheme": "http"}`) require that the value be converted to a keyword to conform to our config schema.  

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/main/com/kubelt/lib/config/util.cljc
+++ b/src/main/com/kubelt/lib/config/util.cljc
@@ -6,28 +6,12 @@
   (:require
    [com.kubelt.lib.config.default :as lib.config.default]))
 
-;; Internal
+;; str->kw
 ;; -----------------------------------------------------------------------------
 
-(defn log-level->keyword
+(defn str->kw
   "Convert a log level string (e.g. 'info', ':info') to a keyword."
   [log-level]
   (if (string? log-level)
-    (keyword (cstr/replace log-level #"^:" ""))
+    (keyword (-> log-level (cstr/trim) (cstr/replace #"^:" "")))
     log-level))
-
-;; Public
-;; -----------------------------------------------------------------------------
-
-#?(:cljs
-   (defn obj->map
-     "Convert a JavaScript configuration object to a Clojure config map."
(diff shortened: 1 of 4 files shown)


src/main/com/kubelt/lib/config/util.cljc
```diff
@@ -24,10 +24,12 @@
...
      [o]
      {:pre [(object? o)]}
-     (let [config  (js->clj o :keywordize-keys true)]
-       (if (empty? config)
-         lib.config.default/sdk
-         ;; Fix up any config map values that didn't translate from
-         ;; JavaScript.
-         (-> config
-             (update-in [:log/level] log-level->keyword))))))
+     (let [config  (js->clj o :keywordize-keys true)
```
tangrammer, line 27: there is a bit of deviation between doc: "Convert a JavaScript configuration object to a Clojure config map." and impl  `(merge lib.config.default/sdk config)`
but also we have 2 definitions with same name and docs but different ns 
https://github.com/kubelt/kubelt/search?q=obj-%3Emap
`com.kubelt.lib.config/obj->map` seems not in use anymore

tangrammer (changes requested):
it seems that obj->map still needs some work

src/main/com/kubelt/lib/config/util.cljc
crimeminister: Good catch on the doc string, I fleshed it out a bit. And I guess our duplicate namespace / fns snuck in during that refactoring pass we did awhile back. I'll reconcile the two.

tangrammer (approved):
LGTM!

PR #482, opened 2022-06-13
chore(sdk): rename p2p to oort

crimeminister:
# Description

Fixes a lingering smell by updating the name used to refer to our backend from an earlier name to the current name (`p2p` to `oort`). 

## Type of change

- [x] Clean-up and refactoring


drew-patel (approved):
Looks good to me! I will need to publish this and test against this on `three-id` but should be pretty straightforward to do. Thanks!

PR #485, opened 2022-06-15
RPC client adaptations and qn_fetchNFTs test

tangrammer:
# Description
There are rpc calls that needs an object instead of an array as params. This PR adds support to derive this type of params based on rpc method spec

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [X] [com.kubelt.rpc.nfts-test](https://github.com/kubelt/kubelt/blob/rpc-adaptations/src/test/com/kubelt/rpc/nfts_test.cljs)

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation website
- [X] I have made corresponding changes to the literal docs
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/src/main/com/kubelt/rpc/request.cljc
+++ b/src/main/com/kubelt/rpc/request.cljc
@@ -59,7 +59,8 @@
         all-params (get method :method.params/all)
         ;; Collect parameters from the supplied parameter map, removing
         ;; any nil entries.
-        params (filter some? (map #(get params %) all-params))]
+        params (cond-> (filter some? (map #(get params %) all-params))
+                 (= "object" (-> method :method/params :params :schema :type)) first)]
     (lib.json/edn->json-forjs-str
      {:id request-id
       :jsonrpc rpc-version
--- a/src/main/com/kubelt/rpc/request.cljc
+++ b/src/main/com/kubelt/rpc/request.cljc
@@ -36,12 +36,13 @@
      :uri/port port
      :uri/path path}))
 
+(def uuid-at (atom 0))
 ;; TODO use an integer stored in client for incrementing counter?
 ;; Or perhaps a ULID / SQUUID to show calling sequence?
 (defn- make-request-id
   "Return a unique request identifier."
   []
-  (lib.uuid/random))
+  (swap! uuid-at inc))
 
 (defn- make-body
   [path method params]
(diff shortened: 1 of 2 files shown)


src/main/com/kubelt/rpc/request.cljc
```diff
@@ -36,12 +36,13 @@
...
      :uri/path path}))
 
+(def uuid-at (atom 0))
 ;; TODO use an integer stored in client for incrementing counter?
 ;; Or perhaps a ULID / SQUUID to show calling sequence?
 (defn- make-request-id
   "Return a unique request identifier."
   []
-  (lib.uuid/random))
+  (swap! uuid-at inc))
```
crimeminister, line 45: Is this something we can do in the backend, maybe by hashing the received request ID before forwarding to a backend service that only accepts integer IDs? JSON-RPC spec says that a string is valid here.

src/main/com/kubelt/rpc/request.cljc
```diff
@@ -36,12 +36,13 @@
      :uri/port port
      :uri/path path}))
 
+(def uuid-at (atom 0))
```
crimeminister, line 39: Would be nice if we could avoid introducing an atom, but if we end up needing it can we keep it in the client map?

src/main/com/kubelt/rpc/request.cljc
crimeminister: @szkl would appreciate your input here!

szkl: The backend can override the request-id for QuickNode. This issue is an example that not all JSON-RPC providers can be faulty. If there is a possible case of using this RPC client against external providers then I guess it'd be helpful to have that capability. Some JSON-RPC client libraries I've checked out accept a function to generate request identifiers for this kind of situation I suppose.

tangrammer: reverted this changes as not necessary by https://github.com/kubelt/kubelt-oort/pull/106

src/main/com/kubelt/rpc/request.cljc
tangrammer: reverted this changes as not necessary by https://github.com/kubelt/kubelt-oort/pull/106

drew-patel (approved):

PR #498, opened 2022-06-28
Store sdk so it could be restored by cljs app

tangrammer:


Cosmin-Parvulescu:
Covered in https://github.com/kubelt/kubelt/pull/509

PR #499, opened 2022-06-28
feat(three-id): deploy to CloudFlare

crimeminister:
# Description

Deploying gating app to CloudFlare.

- [x] update `provider/kubelt.ts` to refer directly to `packages/sdk-web/lib/com.kubelt.sdk.js`
  - if I can figure out a bug I was experience around referring to `@kubelt/sdk-web` instead that would be preferable 
- [x] migrate `packages/three-id` to  `$PROJECT_ROOT/three-id/`
- [x] adds CF proxy worker to `three-id/worker`
- [x] add `three-id/bb.edn` script for building and deploying 
- [x] add `robots.txt` (currently disallows all crawling)
- [x] add `favicon.ico`
  - this isn't being processed by the build and there's already a `favicon.png` so can likely be removed


Cosmin-Parvulescu:
Hmm, does this work with `npm run web`? I got an error locally, expo was trying to load `../../App` and because we now have the structure with `src` it couldn't find the entry point. 

This could be solved with https://docs.expo.dev/guides/monorepos/#change-default-entrypoint but there's also the a bit easier to grok solution where an `App.tsx` is added to the root of the three-id package with the following contents:
```
import App from './src/App'

export default () => <App />
```

crimeminister:
> Hmm, does this work with `npm run web`? I got an error locally, expo was trying to load `../../App` and because we now have the structure with `src` it couldn't find the entry point. Here's a PR https://github.com/kubelt/kubelt/pull/500/files :)

I am not having any trouble running that command on this branch. Please holler if you continue to see any problems.

Cosmin-Parvulescu (approved):
LGTM. Some notes would be the missing version of the Expo npm package.

Cosmin-Parvulescu (approved):

PR #527, opened 2022-07-06
chore(3iD): update build pipeline

crimeminister:
# Description

Updates the `next` pipeline to build and deploy to staging environment.

### DIFF
--- a/.github/workflows/next.yaml
+++ b/.github/workflows/next.yaml
@@ -6,6 +6,10 @@ on:
       - main
 
 jobs:
+
+  # build
+  # ----------------------------------------------------------------------------
+
   build:
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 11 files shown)


.github/workflows/next.yaml
```diff
@@ -29,30 +33,65 @@ jobs:
           node-version: lts/gallium
           cache: 'npm'
           cache-dependency-path: package-lock.json
+
       - run: npm ci
-      - run: npm install cypress --save-dev
+      # - run: npm install cypress --save-dev
```
drew-patel, line 38: I may need this soon-ish when I introduce tests for three-id. Is it cool if I bring it back in at a later date?

drew-patel (approved):

.github/workflows/next.yaml
crimeminister: Yup, let's chat about it when you're ready s.v.p. If we can try to maximize parallelism in the workflows to minimize total runtime, that'd be would be worth a little extra effort as this is our chunkiest job.

PR #533, opened 2022-07-08
BUG(3iD): Polish Tracker - Clear account

Cosmin-Parvulescu:
# Description

> When the user gets to the gate (screen 2C) and clicks "Try Different Wallet" they are sent back to 2A. When they then select Metamask the flow does not re-start. They should be asked to re-connect / re-sign the message.

References https://github.com/kubelt/three-id/issues/92

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [x] Manually

# Checklist:

### DIFF
--- a/three-id/src/provider/web3.ts
+++ b/three-id/src/provider/web3.ts
@@ -9,7 +9,10 @@ const eth = (window as any).ethereum;
 
 export const isMetamask = () => eth?.isMetaMask === true;
 
-export const clearAccount = () => accountSubj.next(null);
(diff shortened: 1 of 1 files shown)


Cosmin-Parvulescu:
Implemented with [this trick](https://github.com/MetaMask/metamask-extension/issues/8990#issuecomment-1154894040). Disclaimers apply. Accounts are not preserved on coming back, however, I've implemented signature re-requesting.

PR #542, opened 2022-07-12
fix(gha): tweak GHA next pipeline

crimeminister:
# Description

Some more pipeline tweaks to make local and remote execution less divergent.


drew-patel (approved):
LGTM!

PR #545, opened 2022-07-12
chore(gha): update release workflow

crimeminister:
# Description

The `release` workflow executes the same steps as for staging, with the exception that the deployment environment is different and the tag name is compiled in as the application version.

### DIFF
--- a/.github/workflows/next.yaml
+++ b/.github/workflows/next.yaml
@@ -14,14 +14,14 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       packages: read
-    services:
-      cloudflare-worker:
-        image: ghcr.io/kubelt/cloudflare-worker:latest
-        credentials:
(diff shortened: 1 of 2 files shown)


.github/workflows/next.yaml
```diff
@@ -14,14 +14,14 @@ jobs:
...
       packages: read
-    services:
-      cloudflare-worker:
-        image: ghcr.io/kubelt/cloudflare-worker:latest
-        credentials:
-          username: ${{ github.actor }}
-          password: ${{ secrets.github_token }}
-        ports:
-          - 8787:8787
+    # services:
```
drew-patel, line 17: Let's remove this all together? I don't see why we need this for now. We can use our devnet and I can remove the artifact too

crimeminister: WFM

PR #562, opened 2022-07-18
Utils to generate ts types from, json-spec based on malli specs

tangrammer:
Relates https://github.com/kubelt/three-id/issues/95

# Description

PoC of generating ts (composed) interfaces/types based on cljs malli specs

## Type of change

# How Has This Been Tested?
So far the utility is called from dev env: 
https://github.com/kubelt/kubelt/blob/json-spec-to-ts/src/dev/jsonspec/utils.cljs#L69

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation website
- [X] I have made corresponding changes to the literal docs
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules


src/dev/jsonspec/utils.cljs
```diff
@@ -0,0 +1,77 @@
...
+   [com.kubelt.lib.promise :as lib.promise]
+   [clojure.string :as str])
+  (:require
+   ["process" :as process]
+   ["json-schema-to-typescript" :refer [compileFromFile]]))
+
+
+(def local-path (str/replace (.cwd process) "packages/sdk-js/lib" ""))
+
+(defn json-spec-path [filename]
```
crimeminister, line 18: Could we generate intermediate files into e.g. the top-level `target/` directory? Or do we need to store them rather than regenerating them during a build?

src/dev/jsonspec/utils.cljs
```diff
@@ -0,0 +1,77 @@
...
+  (str local-path "src/main/com/kubelt/jsonspec/" filename ".json"))
+
+(defn ts-types-path [filename]
+  (str local-path "three-id/src/types/" filename ".d.ts"))
```
crimeminister, line 22: Should we generate these types into `packages/sdk-js` directory for including in package releases?

src/main/com/kubelt/spec/storage.cljc
```diff
@@ -5,14 +5,19 @@
...
+        }
    ;; TODO return error map and the stored value on success?
    [:cat :any] [:or :string :map]])
 
 (def restore-fn
   [:=> {:description "A storage function that returns an SDK instance
 using state loaded from platform storage, or an error map if an error
-occurs."}
+occurs."
+        :json-schema/tsType "(sdk: any) => any"
```
crimeminister, line 19: Does this also require the `json-schema/instanceOf` map entry as for `store-fn`?
```
:json-schema/instanceOf "Function"
```

src/main/com/kubelt/spec/profile.cljc
```diff
@@ -18,12 +18,14 @@
...
-                     [:nickname :string]
-                     [:profilePicture profilePicture]
-                     [:email :string]
-                     [:location :string]
-                     [:job :string]
-                     [:website :string]
-                     [:bio :string]
-                     [:socials socials]]))
+  (mu/optional-keys
+   [:map {:registry {:profile/socials socials
```
crimeminister, line 22: Not sure I understand the need for using `:registry` here. Is it because of the namespaced keyword, or some other reason? i.e. would it work to use:
```clojure
[:profile/profilePicture profilePicture]
```

crimeminister (approved):

src/dev/jsonspec/utils.cljs
tangrammer: yep

src/main/com/kubelt/spec/storage.cljc
tangrammer: yes, comparing current json-schemas generated it seems that we need to be explicit on all json-schema functions

src/main/com/kubelt/spec/profile.cljc
tangrammer: registry is needed to get more than one ts interface from one schema, in the case of Profile i was trying to have same structure as https://github.com/kubelt/kubelt-oort/blob/main/cloudflare-worker/src/packages/profile/types.ts

src/dev/jsonspec/utils.cljs
tangrammer: thinking again, shouldn't be useful that the generated types was also used by oort (as a lib/dep)? if so, perhaps we should define a new build just for types

crimeminister: Perhaps @szkl has an opinion as to how useful it would be to have those types available in the BE?

szkl: I think the versioning would be a problem to be addressed.

szkl: Type mutations might be needed to adapt to the server-side use cases.

szkl: We should look into that with some examples to understand if it can satisfy all programming requirements on the server side.

src/dev/jsonspec/utils.cljs
tangrammer: done!

crimeminister:
Closing for further investigation.

PR #569, opened 2022-07-19
Revert "CHORE(Three-id): Remove manual analytics"

tangrammer:
Reverts kubelt/kubelt#568


alfl (changes requested):
Marking with "Request Changes" to avoid accidental merges.

Cosmin-Parvulescu:
I think we can safely close this :) especially if data is validated already.

PR #588, opened 2022-07-21
chore(sdk): remove some dead code

crimeminister:
# Description

Remove some code that is not currently in use, for great hygiene.


alfl (dismissed):

alfl (approved):
Even more dead code!

PR #608, opened 2022-07-26
chore: smoke testing of internal and external URLs

crimeminister:
# Description

The intent here was to run smoke tests against both internal and external URLs for a given deployment environment. Unfortunately, it seems as though DDoS protection (or a similar capability) kicks in and causes 403 for all URLs after the host name changes from internal to external, or vice versa.

This leaves the changes in place for future refinement, but only execute smoke tests against the *external* URLs for the deployment environment.

### DIFF
--- a/three-id/bb.edn
+++ b/three-id/bb.edn
@@ -848,7 +848,9 @@
                    message (if (= 200 (get response :status))
                              (str "[🪹] Purge zone cache: SUCCESS " details)
                              (str "[❌] Purge zone cache: FAILED " details))]
-               (println message))))}
+               (println message)))
+           ;; Return the set of all URLs that were purged.
+           urls)}
 
   publish:site
   {:doc "Publish assets and worker to CloudFlare"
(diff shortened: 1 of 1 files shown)


three-id/bb.edn
```diff
@@ -937,23 +939,40 @@
...
+  ;; the other.
   test:smoke:http
   {:doc "Perform HTTP request for every deployed asset"
-   :depends [asset:manifest -url:deploy -test:strict?]
-   :task (let [remote-paths (vals asset:manifest)
-               ;; Turn the paths into full URLs.
-               urls (map (partial str -url:deploy) remote-paths)
+   :depends [-urls:internal -urls:external -test:strict? -verbose:level]
+   :task (let [;; Collect all the URLs to be tested.
+               urls (concat -urls:external #_-urls:internal)
```
drew-patel, line 966: Do you want to keep `-urls:internal` ignored?

drew-patel (approved):

PR #609, opened 2022-07-26
chore: update various dependencies

crimeminister:
# Description

Various dependency updates. As always, I am leaving the React (Native) dependencies for expo alone, as it has specific requirements as to version.


three-id/package.json
```diff
@@ -62,10 +62,9 @@
     "@synthetixio/synpress": "1.2.0",
     "@types/react": "17.0.21",
     "@types/react-native": "0.66.13",
-    "expo-cli": "5.5.1",
```
drew-patel, line 65: Was this removed by mistake? Build failure is because `compile:app` needs this dependency.

Screen Shot 2022-07-27 at 09 04 44

crimeminister: Thanks for catching that; I read that new expo uses local binaries rather than requiring a global install of `expo-cli` and was playing around. Looks like I b0rked that dependency.

crimeminister: Re-added `expo-cli` latest released version.

alfl (approved):

PR #637, opened 2022-07-29
docs(claim): Add Authorization Guide and Update 3ID

alfl:
# Description

Add Auth and 3ID docs, with updated RPC names.

## Type of change

- [x] This change requires or is a documentation update

# How Has This Been Tested?

- [x] Local

# Checklist:

- [x] I have made corresponding changes to the documentation website
- [x] I have made corresponding changes to the literal docs


maurerbot (approved):
Approving but we need to change the dir structure of the guide to be top level guides for each major feature (starbase, starport, 3id, orbit). This doc should live under starport guides. Each should also have a a quick start too.

PR #649, opened 2022-08-03
fix(gha): scripts

maurerbot:
Quick fix for gha scripts

### DIFF
--- a/3iD/deploy.edn.example
+++ b/3iD/deploy.edn.example
@@ -56,4 +56,3 @@
  ;; Metamask wallet for 3iD testing.
  :metamask/password ""
  :metamask/recovery-phrase ""}
-c5902365738951d1946e51768978eb46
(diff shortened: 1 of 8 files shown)


3iD/deploy.edn.example
```diff
@@ -0,0 +1,73 @@
...
+;;   [this installs wrangler]
+;; * run: npx wrangler login (in $ROOT/three-id/ directory)
+;;   [grants wrangler permission to make changes in their account]
+;; * ensure they are logged into developer CF site
+;;   NB: in CloudFlare UI make sure you're viewing <dev@kubelt.com>,
+;;   *not* <admin@kubelt.com>
+;; * create worker service: three-id
+;; * create KV store: three-id-APP
+;;
+;; In this file:
```
crimeminister, line 15: Do we want to update these docs? They no longer reflect the process of setting up a new private developer environment.

crimeminister (changes requested):
There are some extra values lingering in `deploy.edn.example`.

PR #722, opened 2022-08-18
fix(3id): check network event

### LINKED ISSUE #651: BUG(gate): Early Access Gate Should Check Network
## I'm submitting a ...

- [x] bug report
- [x] feature request

## What is the current behavior?

Current behaviour is that the early access gate assumes Ethereum mainnet. If your wallet is set to a different chain, or a different network, the flow continues.

## What is the expected behavior?

Expected behaviour is that the user is warned that they should be on Ethereum mainnet to continue.

## What is the motivation / use case for changing the behavior?

It's not always obvious to users when they are on a different chain and it causes confusion, bad UX, and a potential security vulnerability (creates a rare hash collision problem across compatible chains).

poolsar42:
# Description

Now it checks the network in early access gate and also it is checking network change with window.ethereum in onboard. It clears account if network changes from mainnet to something else.

- Closes #651 

## Type of change

### DIFF
--- a/3iD/src/provider/web3.ts
+++ b/3iD/src/provider/web3.ts
@@ -18,23 +18,41 @@ export const getAccountObs = () => {
   return accountSubj.asObservable();
 };
 
+const handleChainChanged = async (network: string) => {
+  if (network) {
+    // We want to be only on ethereum mainnet
+    if (network !== "0x1") {
+      window.alert("Please do not switch the network");
+      clearAccount();
+    }
+    console.warn(`WEB3: Another chain detected`);
+  }
+};
+
 const handleAccountsChanged = async (accounts: string[]) => {
   if (accounts.length > 0) {
     try {
       const provider = await connect();
       const account = await provider.getSigner().getAddress();
-
-      // We want to make sure that the used account
-      // is the same as the signer account;
-      // Getting address via signer will generate
-      // a properly cased address string.
-      if (account.toLowerCase() === accounts[0].toLowerCase()) {
-        // We don't want to force unnecessary refreshes
-        // to subscribers so we don't update
-        // unless the value changed
-        if (account !== accountSubj.getValue()) {
-          accountSubj.next(account);
+      const network = (await provider.getNetwork()).name;
+      // We want to be only on ethereum mainnet
+      if (network === "homestead") {
+        // We want to make sure that the used account
+        // is the same as the signer account;
+        // Getting address via signer will generate
+        // a properly cased address string.
+        if (account.toLowerCase() === accounts[0].toLowerCase()) {
+          // We don't want to force unnecessary refreshes
+          // to subscribers so we don't update
+          // unless the value changed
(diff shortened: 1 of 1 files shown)


3iD/src/provider/web3.ts
```diff
@@ -18,23 +18,41 @@ export const getAccountObs = () => {
   return accountSubj.asObservable();
 };
 
+const handleChainChanged = async (network: string) => {
+  if (network) {
+    // We want to be only on ethereum mainnet
```
maurerbot, line 23: this is not correct. we don't want to be on mainnet only. happy if users want to create account/cores on different networks.

3iD/src/provider/web3.ts
```diff
@@ -18,23 +18,41 @@ export const getAccountObs = () => {
...
-      // a properly cased address string.
-      if (account.toLowerCase() === accounts[0].toLowerCase()) {
-        // We don't want to force unnecessary refreshes
-        // to subscribers so we don't update
-        // unless the value changed
-        if (account !== accountSubj.getValue()) {
-          accountSubj.next(account);
+      const network = (await provider.getNetwork()).name;
+      // We want to be only on ethereum mainnet
+      if (network === "homestead") {
```
maurerbot, line 39: same as before

maurerbot (changes requested):

alfl:
> A network change event probably does count as a logout though.

Not specifically a logout, but a new authenticated SDK for that network, yes. If I change between chains/networks a lot I wouldn't want to get hit with lots of sign-in popups (this does happen if you're using one wallet for Polygon test and main and Ethereum test and main).

@Cosmin-Parvulescu how is save/restore on the SDK looking?

3iD/src/provider/web3.ts
```diff
@@ -18,23 +18,41 @@ export const getAccountObs = () => {
...
+        if (account.toLowerCase() === accounts[0].toLowerCase()) {
+          // We don't want to force unnecessary refreshes
+          // to subscribers so we don't update
+          // unless the value changed
+          if (account !== accountSubj.getValue()) {
+            accountSubj.next(account);
+          }
         }
+      } else {
+        window.alert(
```
alfl, line 53: We'd like to do an in-app alert versus a `window.alert`. @picassobanana this will require some design.

@poolsar42 For now we just want to save and restore (and if necessary re-authenticate) the SDK.

alfl:
@poolsar42 I see what happened here -- the bug report is incorrect. I'll have a 👀 

The desired behaviour is that the user can continue with non-mainnet accounts but we need some logic to handle the state of our client SDK.

alfl (changes requested):

alfl:
Should be https://github.com/kubelt/kubelt/issues/654

3iD/src/provider/web3.ts
poolsar42: from what I see this is exactly what `authenticate` function does

PR #744, opened 2022-08-23
FIX(3iD): Try restore SDK

Cosmin-Parvulescu:
# Description

`getSDK` is now available throughout the client-side code, as such we can't link SDK restoration to it being null. Solution is to force restoration in case of specific `authenticate` call and ignore it if it doesn't return logged in. This should preserve current behavior.

## Type of change

# How Has This Been Tested?

Manually

# Checklist:

### DIFF
--- a/3iD/src/provider/kubelt.ts
+++ b/3iD/src/provider/kubelt.ts
@@ -62,21 +62,21 @@ export const authenticate = async (
   try {
     if (!sdk) {
       sdk = await getSDK();
-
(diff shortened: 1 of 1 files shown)


maurerbot (approved):
fixes reload?

Cosmin-Parvulescu:
> fixes reload?

Nope, it fixes a bug I introduced where restore wasn't called. Now it is called but still SDK answers with error.

PR #753, opened 2022-08-26
fix(3id): double check on nickname

poolsar42:
# Description

When username is not set now it causes an error that View component can't have text element as a child because username returns an empty string. If we add double check this behaviour will change and no error will appear.

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/3iD/package.json
+++ b/3iD/package.json
@@ -60,6 +60,7 @@
     "react-native-screens": "3.11.1",
     "react-native-web": "0.17.7",
     "react-redux": "8.0.2",
+    "react-tooltip": "^4.2.21",
     "redux-persist": "6.0.0",
     "rxjs": "7.5.6",
     "web3modal": "1.9.8",
--- a/3iD/src/screens/AuthLayout.tsx
+++ b/3iD/src/screens/AuthLayout.tsx
@@ -241,7 +241,7 @@ export default function Layout({
           marginVertical: "3em",
         }}
       >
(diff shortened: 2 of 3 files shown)


3iD/package.json
```diff
@@ -60,6 +60,7 @@
     "react-native-screens": "3.11.1",
     "react-native-web": "0.17.7",
     "react-redux": "8.0.2",
+    "react-tooltip": "^4.2.21",
```
Cosmin-Parvulescu, line 63: We use locked package versions, so `^` needs to be removed and afterwards an `npm install` so that package-lock is aligned also. 🙏

3iD/src/screens/AuthLayout.tsx
```diff
@@ -241,7 +241,7 @@ export default function Layout({
           marginVertical: "3em",
         }}
       >
-        {nickname && <SideMenu nickname={nickname} />}
+        {nickname?.length && <SideMenu nickname={nickname} />}
```
Cosmin-Parvulescu, line 244: Can we move this check in SideMenu? Perhaps around [here](https://github.com/kubelt/kubelt/blob/chore/3idpolish/3iD/src/components/side-menu/SideMenu.tsx#L93). Might need to be made optional in the component props.

Cosmin-Parvulescu (changes requested):

PR #869, opened 2022-10-24
feat(sb): work on backend implementation

crimeminister:
# Description

Initial work on Starbase worker, supporting library, and component (durable object providing an OpenRPC interface for remote invocation).


maurerbot (approved):
cc: @szkl

please sync with each other on a commons repo

PR #880, opened 2022-10-26
fix(gha): main publish fails

maurerbot:


projects/oort/webpack.development.ts
```diff
@@ -5,8 +5,8 @@ export default merge(base, {
   mode: 'development',
   devtool: 'inline-source-map',
   output: {
-    filename: 'index.js',
-    libraryTarget: 'commonjs',
+    // filename: 'index.js',
+    // libraryTarget: 'commonjs',
```
szkl, line 9: Why?

projects/oort/wrangler.toml
```diff
@@ -1,26 +1,9 @@
 name = "oort"
 main = "dist/shim.mjs"
 compatibility_date = "2022-09-26"
-workers_dev = true
+# workers_dev = true
```
szkl, line 4: Why?

szkl (approved):

projects/oort/webpack.development.ts
maurerbot: repeated from base

projects/oort/wrangler.toml
maurerbot: using route instead: `route = { pattern = "oort-dev.kubelt.com", custom_domain = true, zone_name = "kubelt.com" }`

PR #902, opened 2022-10-31
chore(icons): set up GHA workflows

crimeminister:
# Description

Set up GitHub Actions workflows for `icons` service.

### DIFF
--- a/.github/workflows/main-icons.yaml
+++ b/.github/workflows/main-icons.yaml
@@ -1,0 +1,49 @@
+# .github/workflows/main-icons.yaml
+---
+name: Icons
+
+on:
+  push:
+    branches:
+      - main
+
+defaults:
+  run:
+    working-directory: ./platform
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: cachix/install-nix-action@v18
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+
+      - run: nix-build platform.nix
+
+      - name: Install
+        run: yarn workspaces focus -A
+
+      # - name: Test
+      #  run: yarn workspaces foreach -i -v -t --include icons run test
+
+      - name: Deploy Icons to Dev Worker
+        uses: cloudflare/wrangler-action@2.0.0
+        with:
+          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          workingDirectory: "platform/icons"
+          command: publish --env dev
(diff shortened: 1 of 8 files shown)


platform/icons/wrangler.toml
```diff
@@ -14,21 +14,17 @@ route = { pattern = "icons-dev.kubelt.com/*", zone_name = "kubelt.com" }
...
+# npx wrangler --var ACCOUNT_ID=<...>
 
 [env.dev]
 name = "icons-dev"
 route = { pattern = "icons-dev.kubelt.com/*", zone_name = "kubelt.com" }
 
 [env.dev.vars]
 # The account ID that owns uploaded images.
-ACCOUNT_ID = "2906425795180ede5ac707178f96406d"
+ACCOUNT_ID = ""
```
maurerbot, line 27: You can probably delete these completely?

maurerbot (changes requested):

platform/icons/wrangler.toml
crimeminister: Just left them in as a reminder to set a value using `--var` when running locally, unless you have that sorted in the workspace scripts somehow?

crimeminister: The reason this might need to work a bit differently is that this service requires the `ACCOUNT_ID` value at runtime to construct signed URLs, having that value available is not just to allow wrangler to publish the service.

szkl (approved):

.github/workflows/main-icons.yaml
```diff
@@ -0,0 +1,43 @@
+# .github/workflows/main-icons.yaml
```
maurerbot, line 1: we should add this badge to the root readme

maurerbot (changes requested):

platform/icons/wrangler.toml
crimeminister: It's available in the GHA build context (as we set it in the workflow), but is it available at runtime when the worker is executing if we don't specify as part of the `wrangler.toml`? I am perusing the docs to see what the default environment contains, if anything.

.github/workflows/main-icons.yaml
crimeminister: Updated.

.gitignore
```diff
@@ -118,6 +118,8 @@ web_modules/
 .env.production.local
 .env.local
 
+.dev.vars
```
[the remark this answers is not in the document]
maurerbot: 👍

maurerbot (approved):

PR #908, opened 2022-11-01
FEAT(3iD): Expand providers

Cosmin-Parvulescu:
# Description

The public provider is overused and ends up throwing 429 errors on and on. The solution alchemy proposes is a retry strategy, which is exactly what WAGMI is doing with it. But that doesn't really work as the public endpoint, especially for Goerli is heavily used. As such, we need to add our own Alchemy API key in an Alchemy Provider and use that in case of throttling. 

Both providers are now in use, but in real life scenarios, the public one fails constantly while we can rely on the AlchemyProvider to get an actual response once NFTs are minted.

I attempted to load the environment variables in client only by checking for the presence of `window` but I ran into a lot of troubles and decided to go ahead with server sent secrets in the loader.

Also updated GHA to contain the secrets.

## Type of change

# How Has This Been Tested?

I have ran the flow on multiple accounts.

# Checklist:


apps/threeid/.dev.vars.example
```diff
@@ -35,3 +35,6 @@ NFTAR_CHAIN_ID=5
 VALID_CHAIN_ID_NAME=goerli
 
 ALCHEMY_NFT_API_URL=
+ALCHEMY_PUBLIC_API_KEY=
+
+TEST_ENV=true
```
maurerbot, line 40: this should be false by default

.github/workflows/release-threeid.yaml
```diff
@@ -68,6 +69,7 @@ jobs:
           MINTPFP_CONTRACT_ADDRESS: ${{ secrets.MINTPFP_CONTRACT_ADDRESS }}
           NFTAR_CHAIN_ID: 1
           VALID_CHAIN_ID_NAME: mainnet
+          ALCHEMY_PUBLIC_API_KEY: ${{ secrets.ALCHEMY_PUBLIC_API_KEY }}
```
maurerbot, line 72: is this better than than using the url?

maurerbot (changes requested):

apps/threeid/.dev.vars.example
Cosmin-Parvulescu: Updated

PR #913, opened 2022-11-02
fix(remix): side effects issue

maurerbot:
# Description

Fixes local wrangler requiring side effects on and service bindings wrapped up.

## Type of change

# How Has This Been Tested?

Local wrangler


Cosmin-Parvulescu (approved):
We should find the time to setup prettier and / or linting so we don't rely on individual IDE configurations.

Cosmin-Parvulescu (changes requested):
ThreeID Build failing & Checks not passing.

Cosmin-Parvulescu (approved):

PR #951, opened 2022-11-07
fix(sb): use yarn workspace resolution over file:// deps

crimeminister:
# Description

Replaces local `file://` package dependency specification with yarn workspaces resolution across monorepo packages with the hope that this unbreaks the build.


szkl (approved):

szkl (changes requested):
Need to update `yarn.lock` too.

szkl (approved):

PR #956, opened 2022-11-07
chore(workspaces): update workspace configuration

szkl:


szkl:
Another pull request later would be time-saving right now.

maurerbot (approved):

PR #994, opened 2022-11-09
chore(sb): split starbase into separate packages and platform worker

crimeminister:
# Description

Move things around to better conform to new repository layout. In particular:
- [x] `platform/starbase/worker` → `platform/starbase`
- [x] `platform/starbase/components` → `packages/cores`
- [x] `platform/starbase/lib/openrpc` → `packages/openrpc`

Along the way I update the `package.json` of the affected packages to include:
- `prettier`
- `eslintrc`
- `eslint-plugin-tsdoc`

and associated script aliases, as well as updating `.eslintrc.json` to enable `tsdoc` lint warnings about mal-formatted comments.

NB: this deletes `platform/yarn.lock`. This file was preventing cross-repo/workspace dependency linking.


crimeminister:
@maurerbot Should there not be just a single top-level `yarn.lock`?

crimeminister:
> 

FWIW, it was only after I removed this file that I was able to rely on something in `packages/cores` from `platform/starbase`. Before that, running `npx yarn install` in `platform/starbase` would result in a missing dependency error.

crimeminister:
@maurerbot Sounds like we have agreement to rework the workflow to expect a single top-level `yarn.lock`. Work for you?

maurerbot (approved):

PR #1005, opened 2022-11-10
chore(sb): workflows

crimeminister:
# Description

Adds GHA workflows for starbase.

- `main-starbase`
- `next-starbase`
- `release-starbase`
- sets up service binding for `console-{dev,next,current}` to `starbase-{dev,next,current}`

## note

I did some experimentation with dependency caching; the `actions/cache@v3` action doesn't appear to care about the default `working-directory` setting, relative paths in the `path:` configuration and `hashFiles()` argument result in an error message.

### DIFF
--- a/apps/console/wrangler.toml
+++ b/apps/console/wrangler.toml
@@ -46,7 +46,6 @@ local_protocol = "http"
 # -----------------------------------------------------------------------------
 
 [env.dev]
-name = "console-dev"
 
 services = [
   { binding = "STARBASE", service = "starbase-dev", environment = "production" }
@@ -70,7 +69,6 @@ STORAGE_NAMESPACE = "starbase"
 # -----------------------------------------------------------------------------
 
 [env.next]
-name = "console-next"
 
 services = [
   { binding = "STARBASE", service = "starbase-next", environment = "production" }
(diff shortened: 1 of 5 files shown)


platform/starbase/wrangler.toml
```diff
@@ -33,3 +41,45 @@ new_classes = [
...
+  { name = "STARBASE_USER", class_name = "StarbaseUser" },
+]
+
+[env.dev.vars]
+
+# Environment: next
+# -----------------------------------------------------------------------------
+
+[env.next]
+name = "starbase-next"
```
maurerbot, line 63: we can leave names out. --env <env> names the worker

maurerbot (changes requested):

platform/starbase/wrangler.toml
crimeminister: Fixed.

apps/console/wrangler.toml
```diff
@@ -92,9 +95,15 @@ STORAGE_NAMESPACE = "starbase"
 
 [env.current]
 name = "console-current"
```
maurerbot, line 97: need to remove name here too

maurerbot (changes requested):

apps/console/wrangler.toml
crimeminister: It's removed with last commit.

maurerbot (approved):
We can delete workers in CF that will be orphaned

PR #1027, opened 2022-11-14
feat(sb): bind to passport service to authenticate calls

crimeminister:
# Description

Use `@kubelt/platform.commons` utility `isAuthenticated()` from starbase worker to authenticate requests.

Incidental:
- remove a couple of legacy dependencies from root `package.json`:
  - `classnames`: not used with tailwind
  - `wagmi`: should be required by specific workspaces that require it

### DIFF
--- a/platform/starbase/wrangler.toml
+++ b/platform/starbase/wrangler.toml
@@ -14,11 +14,11 @@ workers_dev = false
 
 # Services
 # -----------------------------------------------------------------------------
-
 # A list of other Clouflare services bound to this service.
-# services = [
-#   { binding = "AUTH", service = "auth", environment = "production" },
-# ]
+
+services = [
+  { binding = "Account", service = "account" }
+]
 
 # Secrets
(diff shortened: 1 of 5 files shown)


platform/starbase/wrangler.toml
```diff
@@ -14,11 +14,11 @@ workers_dev = false
...
 # Services
 # -----------------------------------------------------------------------------
-
 # A list of other Clouflare services bound to this service.
-# services = [
-#   { binding = "AUTH", service = "auth", environment = "production" },
-# ]
+
+services = [
+  { binding = "PASSPORT", service = "passport-dev" }
```
szkl, line 20: Binding name is [`Passport`](https://github.com/kubelt/kubelt/blob/main/platform/worker-commons/src/utils/isAuthenticated.ts#L10).

szkl (changes requested):

szkl (approved):

PR #1045, opened 2022-11-17
feat(sb): misc improvements

crimeminister:
# Description

- [x] make auth checks fine-grained
  - when creating an `openrpc.method()` an `auth` fn can be supplied
  - this is effectively a middleware that is executed for that specific RPC method invocation
  - if it returns `void`, the auth check has succeeded
  - if it returns a `Response`, the request handling is short-circuited and the response is returned
- [x] wraps the commons `isAuthenticated()` check in a middleware for `platform/starbase`
- [x] applies the auth checking for every `starbase` service method except for:
  - `kb_initPlatform`
  - `kb_appProfile`
  -  `kb_appAuthCheck`
- [x] adds skeleton methods:
  - `kb_appRotateSecret`
  - `kb_appPublish`
  - `kb_appAuthCheck` (replaces `kb_appAuthInfo`)
- [x] initial work on app indexing for listing and lookup

### DIFF
--- a/packages/cores/application/index.ts
+++ b/packages/cores/application/index.ts
@@ -3,6 +3,8 @@
  * @file src/index.ts
  */
 
+import * as _ from "lodash";
+
 import * as openrpc from "@kubelt/openrpc";
 
 import type {
@@ -128,10 +130,33 @@ export class StarbaseApplication {
   ): Promise<RpcResult> {
     const app = input.get("app");
 
-    return Promise.resolve({
-      invoked: "app_fetch",
-      app,
-    });
+    return Promise.resolve(app);
+  }
+
+  // public_profile
+  // -----------------------------------------------------------------------------
+
+  @method("public_profile")
+  @requiredScope("starbase.read")
+  @requiredField("app", [FieldAccess.Read])
+  publicProfile(
+    params: RpcParams,
+    input: RpcInput,
+    output: RpcOutput,
+  ): Promise<RpcResult> {
+    const app = input.get("app")
+
+    // If the application is not published we shouldn't return any
+    // information.
+    let profile = {}
+    if (app?.published == true) {
+      profile = _.omit(app, [
+        'clientSecret',
+        'published',
+      ])
+    }
+
+    return Promise.resolve(profile)
   }
 
   // ---------------------------------------------------------------------------
--- a/packages/cores/application/schema.ts
+++ b/packages/cores/application/schema.ts
@@ -38,6 +38,19 @@ const rpcSchema: RpcSchema = {
       },
       errors: [],
     },
+    {
+      name: "public_profile",
+      summary: "Return public view of application data",
+      params: [],
+      result: {
+        name: "profile",
+        description: "Public application data",
+        schema: {
+          type: "object"
+        }
+      },
+      errors: [],
+    },
   ],
   components: {
     contentDescriptors: {
--- a/packages/cores/user/index.ts
+++ b/packages/cores/user/index.ts
@@ -51,6 +51,11 @@ import schema from "./schema";
   validator: (x) => { return true },
   */
 })
+@field({
+  name: "index",
+  doc: "An index from app attributes into app IDs",
+  defaultValue: {},
+})
 export class StarbaseUser {
 
   // user_name
@@ -70,27 +75,6 @@ export class StarbaseUser {
     });
   }
 
-  // add_application
-  // ---------------------------------------------------------------------------
-
-  @method("add_application")
-  @requiredScope("starbase.user")
-  @requiredField("apps", [FieldAccess.Read, FieldAccess.Write])
-  addApplication(
-    params: RpcParams,
-    input: RpcInput,
-    output: RpcOutput,
-  ): Promise<RpcResult> {
-    const appSet = input.get("apps")
-    const appId = params.get("appId")
-    // Add the supplied application ID to the list of user's applications.
-    if (appId !== undefined) {
-      appSet.add(appId)
-      output.set("apps", appSet)
-    }
-    return Promise.resolve(appId);
-  }
-
   // list_applications
   // ---------------------------------------------------------------------------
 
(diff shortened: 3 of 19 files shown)


platform/starbase/curl/kb_appAuthCheck.json
```diff
@@ -0,0 +1,12 @@
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "kb_appAuthCheck",
+  "params": {
```
crimeminister, line 5: @szkl I'd appreciate it if you could help me populate this with example request data.

packages/cores/application/index.ts
```diff
@@ -134,6 +136,32 @@ export class StarbaseApplication {
...
+    params: RpcParams,
+    input: RpcInput,
+    output: RpcOutput,
+  ): Promise<RpcResult> {
+    const app = input.get("app")
+
+    // If the application is not published we shouldn't return any
+    // information.
+    let profile = {}
+    if (app?.published == true) {
```
maurerbot, line 155: shouldn't we return an error code?

packages/cores/application/schema.ts
```diff
@@ -38,6 +38,19 @@ const rpcSchema: RpcSchema = {
       },
       errors: [],
     },
+    {
+      name: "public_profile",
```
maurerbot, line 42: maybe just "app_profile"? or "public_app_profile"?

packages/cores/user/index.ts
```diff
@@ -51,6 +51,11 @@ import schema from "./schema";
   validator: (x) => { return true },
   */
 })
+@field({
+  name: "index",
+  doc: "An index from app attributes into app IDs",
+  defaultValue: {},
+})
 export class StarbaseUser {
```
maurerbot, line 59: not sure we need this object at all. why not just create an edge between a 3ID account and app?


TBD

packages/cores/application/index.ts
crimeminister: We can return an error from the worker if that's what you want the caller to get?

packages/cores/application/schema.ts
crimeminister: I was actually thinking of reducing the name to just `profile` since it's a method on an application already, and this name is turned into a method on the constructed client, e.g. `app.appProfile()` vs. `app.publicAppProfile()` vs `app.profile()`. These names are never exposed outside of a worker, though, so it's more for our internal consumption and preference for naming scheme.

packages/cores/user/index.ts
crimeminister: That's what I'm planning to do, but it's not ready yet (given what we talked about re. storing edges in a DB, etc.) so this is mean to be a stopgap. The other concern is what to do with console-specific information about a user, e.g. if a user has "settings" for console (number of images to show in a page, etc.), I would default to putting it in this object to keep things encapsulated. Alternately, I could store that kind of data in a remote service e.g. StorageAPI, or would need a way to associate arbitrary info with a User core.

packages/cores/application/index.ts
maurerbot: why not bubble it up from here? if asking for the profile through the public method and it's not published I'd expect a 404

packages/cores/application/schema.ts
maurerbot: would `app.profile()`  then handle auth requests too and return secrets if present?

packages/cores/user/index.ts
maurerbot: If next steps is to start setting up edges and we haven't opened up console to the public yet we can likely forego this object. For storing settings we could possibly use the storage api in the user account object d1, or something else (TBD how later)

packages/cores/application/index.ts
szkl: Do you mean HTTP 404 or an "not found" error as a JSON-RPC response?

platform/starbase/curl/kb_appAuthCheck.json
```diff
@@ -0,0 +1,12 @@
...
+    "appId": "@kubelt/console",
+    "redirectURL": "https://console.kubelt.com/dashboard",
```
szkl, line 7: Shouldn't this be called `redirectUri` as it might not be a location depending on the application?

platform/starbase/curl/kb_appAuthCheck.json
szkl: It looks good as it is. Just needs some scopes.

platform/starbase/curl/kb_appAuthCheck.json
```diff
@@ -0,0 +1,12 @@
...
+    "scopes": [],
```
szkl, line 8:
```suggestion
    "scopes": ["app:admin"],
```

platform/starbase/curl/kb_appAuthCheck.json
szkl, line 7:
```suggestion
    "redirectUri": "https://console.kubelt.com/dashboard",
```

szkl (changes requested):

packages/cores/application/schema.ts
crimeminister: `kb_appProfile` is meant to just return public profile information. The complete app record is available from `kb_appFetch`. And after our discussion yesterday I have added `kb_appAuthCheck` to perform a check as part of the auth flow.

packages/cores/user/index.ts
crimeminister: My immediate need is to store a list of user's apps so I can display those options in the console UI. Do you want that info in the storage API, accessed by service binding to oort?

I had planned to create the user object and others as well, e.g. a node for each smart contract belonging to an app (associated with links), and those apps linked to a local user node, and the local starbase user node linked to the canonical user core owned by oort that identifies each user uniquely. Should we be leaning into the graph architectural pattern, multiplying the number of nodes and edges we create, keeping nodes small (storing only locally useful data) and mediating access to the data via scopes, etc.?

platform/starbase/curl/kb_appAuthCheck.json
crimeminister: I changed it to match what was already stored, and because I wasn't imagining a redirect to anything but a web app. Happy to change it to be `Uri` if you think that fits better?

packages/cores/user/index.ts
maurerbot: If it's just a list of users apps use setup a starbase d1, there isn't a reason we need a temporary solution for this rn. If we need to stor user setting for a specific application we should probably use the users storage api and setup a namespace for starbase but, if this isn't going to work we should then use d1.

In the future, if we do anything quick and temporary KVs would be preferable to DOs. We don't need to think to hard about abstractions for first versions.

platform/starbase/curl/kb_appAuthCheck.json
maurerbot: I don't think this one needs scopes.

packages/cores/application/index.ts
crimeminister: > 

Where possible, I am trying to deal with exceptional conditions and errors in the workers to avoid complicating DOs, which I see mainly as simple state management and synchronization primitives.

packages/cores/user/index.ts
crimeminister: Sorted this out offline. Thanks for the feedback @maurerbot.

platform/starbase/curl/kb_appAuthCheck.json
crimeminister: @szkl I committed your change adding an example scope. If it turns out we don't need that info as part of the auth flow, please let me know and I'll update the list of parameters for this call in the schema and avoid including scopes as part of the comparison check.

maurerbot (approved):

maurerbot (approved):

szkl (approved):

PR #1051, opened 2022-11-18
feat(sb): add method kb_appCreate to create OAuth app profile

crimeminister:
# Description

## StarbaseApplication

- [x] add `init` method to set initial data for the component
- [x] rename `app_fetch` method to `fetch`
- [x] rename `public_profile` method to `profile`
- [x] rename `store` method to `update`
  - only a "safe" subset of the app data can be modified (the public profile parts)

## OpenRPC client

- [x] fix bug in DO addressing
  - add `id` and `name` parameters to `openrpc.discover()` options
  - if an ID is supplied, it is treated as a serialized DurableObjectId
  - if a name is supplied, the object ID is derived from the given string
  - if neither is supplied, a new random ID is generated instead
  - the ID of the component (DO, core) the client is for can be retrieved as `stub.$.id`

## Starbase

- [x] stores a mapping from application OAuth client ID to internal application ID
  - kept in the `FIXTURES` KV store
  - all service methods that need to identify an application should now do so using the client ID
- [x] method `kb_appAuthId` now identifies app from OAuth client ID
  - removed the explicit `appId` field from example request `curl/kb_appAuthCheck.json`  
- [x] add `kb_appCreate` RPC method to `platform/starbase`
  - this returns: `clientId`, `clientSecret`
  - the client secret is stored hashed, and is never available in plaintext after this call returns
  - all API calls that interact with an application core identify that application by its OAuth client ID
- [x] added method `kb_appUpdate` to allow for updating of public app profile data
- [x] remove service binding to `account` service
  - pending changes in auth flow make the binding superfluous

## Dependencies

- [x] bump `wrangler` to `2.4.2`
- [x] bump `actions/checkout` to `v3` (in `pr-packages` workflow)
- [x] bump `@cloudflare/workers-types` to `3.18.0`


platform/starbase/src/index.ts
```diff
@@ -138,8 +141,42 @@ const kb_appCreate = openrpc.method(schema, {
...
+      // TODO better type for JWT?
+      const token: string = context.get(KEY_TOKEN)
+
+      const clientId = oauth.makeClientId()
+      const clientSecret = oauth.makeClientSecret()
+      const hashedSecret = await secret.hash(clientSecret)
+
+      const app = await openrpc.discover(sbApplication, undefined, {
+        token,
+        tag: 'stabase-app'
```
szkl, line 155: typo

szkl (approved):

platform/starbase/src/index.ts
crimeminister: Fixed! Thanks.

crimeminister: Actually, what I'd like is to extract the name of DO from the binding automagically, if that's possible. 🤔

crimeminister:
@Cosmin-Parvulescu This ends up touching some deps in `threeid`. I think the changes are innocuous, but still wanted to call your attention to it.

maurerbot (approved):

PR #1079, opened 2022-11-24
feat(address): generate URNs for `address` and `coreId` properties

szkl:


platform/address/src/core.ts
```diff
@@ -23,7 +24,20 @@ export default class Core extends DurableObject<Environment, CoreApi> {
 
   async setAddress(address: string, coreId: string): Promise<void> {
     const type = getType(address)
-    await this.storage.put({ address, type, coreId })
+    if (!type) {
+      throw 'unsupported type'
+    }
+
+    await this.storage.put({
+      address: URN.generateUrn('address', 'threeid.xyz', type, 'name', address),
```
maurerbot, line 32: type as query param as dicussed (prob needs a urn package update)

platform/address/src/core.ts
```diff
@@ -23,7 +24,27 @@ export default class Core extends DurableObject<Environment, CoreApi> {
...
+    }
+
+    await this.storage.put({
+      type,
+      address: URN.generateUrn(
+        'address',
+        'threeid.xyz',
+        'address',
+        'name',
+        address
```
maurerbot, line 38: in next PR when we update this package we can make the last param a map of descriptors

maurerbot (approved):

PR #1083, opened 2022-11-24
chore(deps): bump eslint to 8.28.0

crimeminister:
# Description

Use `eslint 8,28.0` consistently across the repo.

- [x] set the `--immutable` flag on the `yarn install` command in `build-test-apps` workflow
  - this ensures that the lock file isn't modified during the install
  - avoid setting `--immutable-cache` as the GitHub cache may need updating, e.g. because of removed dependencies


szkl (commented):
> this ensures that the lock file isn't modified during the install

Shouldn't it fail if the lock file needs to be updated? Is there an earlier job that checks it?

maurerbot (approved):

PR #1085, opened 2022-11-25
feat(threeid): update react

Cosmin-Parvulescu:
# Description

Increased ThreeID React major version (18+)

Added Spinner to design system
Added Cover component with outside controlled state to design system
Added Storybook to ThreeID
Added Cover with hooks to ThreeID Storybook

## Type of change

- This change requires or is a documentation update

# How Has This Been Tested?

Painfully

# Checklist:


crimeminister:
Does it make sense to update React across the repo? Do we want to try and stick with the same version of major deps like React when that's possible?

apps/threeid/.storybook/preview.js
```diff
@@ -0,0 +1,12 @@
+import '../app/styles/tailwind.css'
+import '@kubelt/design-system/dist/index.css'
```
maurerbot, line 2: let's use the src

maurerbot (changes requested):

Cosmin-Parvulescu:
> Does it make sense to update React across the repo? Do we want to try and stick with the same version of major deps like React when that's possible?

Right now only solution not running 8.2 is Passport. I think that should be done as part of Adrian's changes or another PR maybe. :)

apps/threeid/.storybook/preview.js
Cosmin-Parvulescu: Updated

maurerbot (approved):

PR #1089, opened 2022-11-28
feat(platform): new urn and address objects

maurerbot:
# Description

Setup new URN and add specialized cyrpto address core

## Type of change

# How Has This Been Tested?

wrangler


platform/address/src/utils.ts
```diff
@@ -1,46 +1,50 @@
...
-      }
-      address = resolvedType.address
-      break
-    }
-    default:
-      throw `unsupported address type ${type}`
+  if (!fragment) {
+    throw new Error(
+      `missing 3RN fragment. Expected "<core type>:<address type>"`
+    )
```
szkl, line 39: Why "core type:address type" is the fragment? They are more like `r-component`.

> The r-component is intended for passing parameters to URN resolution
   services (taken broadly, see [Section 1.2](https://www.rfc-editor.org/rfc/rfc8141.html#section-1.2)) and interpreted by those
   services.  (By contrast, passing parameters to the resources
   identified by a URN, or to applications that manage such resources,
   is handled by q-components as described in the next section.)

Interpreting the types is something that happens before the object is located so they are not a query parameter to the object.

> The f-component is intended to be interpreted by the client as a
   specification for a location within, or region of, the named
   resource.

These types are taking place in the process of locating a durable object so I think they are actually a `r-component`.

platform/address/src/utils.ts
```diff
@@ -1,46 +1,50 @@
...
+}> => {
+  const urn = request.headers.get('X-Resource-3RN') as AddressURN
+  // urn:threeid:address:?+<rcomponent>?=name=<do from name>#<do object class>:<class subtype>
+
   if (!urn) {
     throw new Error('missing X-Resource-3RN header')
   }
-  const { service, domain, object, descriptors } = parseUrn(urn)
+
+  const { nid: domain, nss: service, qcomponent, fragment } = parseURN(urn)
```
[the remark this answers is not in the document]
maurerbot: thinking this is a dedicated method within the address service. romponents are relay info only and can be ignored?

platform/address/src/utils.ts
maurerbot: Using fragment as local information where local is the DO.

RN the interpreation of rcomponent is for relay only (aka URN resolution services)

maurerbot: kind of convenient r for relay

maurerbot: a decent change would be to leave frgment as `#eth` and push `type=cypto` to q component? @szkl

platform/address/src/utils.ts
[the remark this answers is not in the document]
szkl: I don't think we should designate an element in the URN to a specific service layer and that's I guess my main argument in the discussion below.

platform/address/src/utils.ts
szkl: but this information is referenced and used before there is a DO instance, in the worker process. Hence, that's why I think they are parameters for the URN resolver instead of a part of the resource.

szkl (changes requested):
It looks like `Core` is actually an abstract class definition and there won't be an instance of it ever. Perhaps declare it an abstract class and remove its durable object binding?

The type declarations related to `CryptoCore` should extend the base types.

platform/address/src/jsonrpc.ts
```diff
@@ -25,57 +24,78 @@ import {
   CryptoCoreApi,
   WorkerApi,
   CryptoWorkerApi,
-  CryptoAddressType,
   AddressProfile,
+  CryptoCoreType,
 } from './types'
 import { resolve3RN } from './utils'
+import { default as CryptoCoreStatic } from './crypto-core'
```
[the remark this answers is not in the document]
maurerbot: so we can call static methods on the core before init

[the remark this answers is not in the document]
maurerbot: this won't work because the default is CryptoCore and that is used in the Env binding

PR #1091, opened 2022-11-28
fix(passport): urns

maurerbot:


szkl (approved):

platform/address/wrangler.toml
```diff
@@ -5,7 +5,9 @@ node_compat = true
 workers_dev = false
 
 durable_objects.bindings = [
-  { name = "Core", class_name = "Core" }
+  { name = "Core", class_name = "Core" },
```
szkl, line 8: Is there a case where a `Core` is created?

maurerbot: Not at the moment

PR #1104, opened 2022-11-30
wip(sb): graph package for edges

maurerbot:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

- Closes # (issue)
- Closes # (issue)
- Closes # (issue)

## Type of change

- This change requires or is a documentation update
- Security Fix

# How Has This Been Tested?

# Checklist:


packages/openrpc/impl/client.ts
```diff
@@ -47,8 +53,16 @@ interface RpcDispatch {
 
 // A map of data properties we expose with the .$ accessor.
 type RpcProperties = {
```
crimeminister, line 55: @szkl This was the change I mentioned, adding the ability to specify the DO to create a client for using the URN rather than just the ID (string-serialized DurableObjectId) or name (arbitrary string).

szkl: I think it'd be practical to start with constructing URNs for logical entities. The lower-level entities like durable objects in general could be represented as a URN too but I don't think it is worth implementing a process that handles it right now as the durable object applications can depend on various conditions and that could prove a higher development cost than it is necessary. If the logical entity can be represented as a durable object then the URN would be enough to `DurableObjectNamespace.idFromName()`. Though, the contents of the URN shouldn't be tied to DO or any other implementation detail.

szkl: I'm still developing ideas to establish a practical format using the `urns` package so I guess I can share some examples once that's finalized.

maurerbot (commented):
needs GHA

crimeminister:
> needs GHA



> needs GHA

The tests are already running, so will add GHA for deploying the worker. Do you want to reset the `dev` and `next` database on each release at the same time, dropping any accumulated data?

crimeminister:
> needs GHA

First pass GHA done. As discussed elsewhere, we don't drop the database for `next`.

Issue #1126, opened 2022-12-03
feat(console): create application


maurerbot:
# Why

Developers need to create apps in console

# What

A dashboard for listing applications, create application wizard, a dashboard for the application and editing 0xAuth

# How

- [ ] Console Dashboard and nav
- [ ] Create App
- [ ] App Dashboard
- [ ] App 0xAuth Setup 

# Notes

- https://www.figma.com/file/EqUEbCGHGnZXDMSUbrxhX7/Kubelt?node-id=4070%3A12538&t=THGcHuV3uJsm5LeM-0
- https://www.figma.com/file/EqUEbCGHGnZXDMSUbrxhX7/Kubelt?node-id=5283%3A14983&t=THGcHuV3uJsm5LeM-0
- https://www.figma.com/file/EqUEbCGHGnZXDMSUbrxhX7/Kubelt?node-id=4359%3A16220&t=THGcHuV3uJsm5LeM-0

PR #1137, opened 2022-12-06
feat(starbase): create app closes: #1126

maurerbot:
Big PR that kicks up edges integration into console


platform/starbase/src/index.ts
```diff
@@ -11,6 +11,8 @@ import * as _ from 'lodash'
 
 import * as set from 'ts-set-utils'
 
+import * as graph from '@kubelt/graph'
```
crimeminister, line 14: I am calling a couple of utility methods from here.

platform/starbase/src/index.ts
```diff
@@ -115,16 +115,31 @@ const authCheck: RpcAuthHandler = async (
     throw new Error("missing environment; can't perform auth check")
   }
 
-  // Create a version of isAuthenticated() that doesn't require entire
-  // Env, just the specific service binding proxy for Account?
-  /*
+  // Perform a request to the Access service, passing it the token
```
crimeminister, line 118: This updates the auth checking fn to call the `access` service, passing along the JWT received with the current request. This fn is invoked for the methods that need authorization; this is indicated by setting this auth fn as the argument to the `auth` config property when defining an RPC method handler.

platform/starbase/wrangler.toml
```diff
@@ -37,13 +37,6 @@ kv_namespaces = [
   { binding = "LOOKUP", id = "27035c39f40343a4a0deab740f2b9fcc", preview_id = "883765b4372d42bea90c67e2e6bfb64b" }
 ]
 
-# D1 Databases
```
crimeminister, line 40: Remove D1 binding since we're talking to the edges service to create, remove, or find edges instead. That service is the only thing that needs to bind the database in this pattern.

platform/starbase/wrangler.toml
```diff
@@ -64,18 +57,17 @@ port = 9090
...
+# new_classes = [
+#   "StarbaseApplication",
+# ]
 # renamed_classes = [
 #   { from = "OldName", to = "NewName" }
 # ]
-# deleted_classes = [
-#   "DeprecatedClass"
-# ]
+deleted_classes = [
```
crimeminister, line 67: Add a migration that removes instances of the now-defunct `StarbaseContract` and `StarbaseUser` durable objects.

platform/starbase/wrangler.toml
```diff
@@ -101,12 +93,12 @@ new_classes = [
 [env.dev]
 
 services = [
+  { binding = "ACCESS", service = "access-dev" },
```
crimeminister, line 96: Add service bindings to the `access` service (for verifying auth tokens), and the `edges` service (for working with edges).

platform/starbase/wrangler.toml
```diff
@@ -138,17 +130,19 @@ kv_namespaces = [
 
 [env.next.vars]
 ENVIRONMENT = "next"
-PLATFORM_OWNER = "@kubelt"
```
crimeminister, line 141: This was a hard-coded name for the user that owns the application nodes for our internal platform apps. I imagine that we'll create a DO for our "internal" user, and link the app nodes for `console`, `threeid`, etc. to that owner node instead?

platform/starbase/src/index.ts
```diff
@@ -11,11 +11,13 @@ import * as _ from 'lodash'
 
 import * as set from 'ts-set-utils'
 
+import * as graph from '@kubelt/graph'
+
 import * as openrpc from '@kubelt/openrpc'
 
 import invariant from 'tiny-invariant'
 
-import checkEnv from '@kubelt/platform.commons/src/utils/checkEnv'
```
crimeminister, line 18: Removing usage of `checkEnv` utility, it throws an error if there is no `Account` binding in the worker environment. Since that's no longer what we use for auth checks this shouldn't be necessary (but it might be a good idea to update it to check for an `ACCESS` service binding instead).

crimeminister (approved):

PR #1175, opened 2022-12-08
feat(openrpc): add support for scheduled alarms on components

crimeminister:
# Description

Add support for Durable Object alarms to the `@kubelt/openrpc/components` package.

- [x] add `RpcAlarm` utility class to allow for alarm times to be set
  - `RpcAlarm.at('absolute time')`
  - `RpcAlarm.after(`duration`)`
- [x] injects alarm utility into RPC method handlers on a component
- [x] inspects alarm instance after method is invoked and sets alarm to returned timestamp (if defined)
- [x] allow a method to be marked with `@alarm` decoration (instead of `@method`)
  - this should trigger the invocation of that method when the alarm fires
- [x] allow alarm method to declare fields access the same as methods
- [x] inject field values as inputs when invoking alarm method
- [x] store outputs returned by the alarm method
- [x] throws an error if user tries to mark more than one method with `@alarm`

Adds some `cURL` fixture data for testing:
- [x] `kb_init` - a JSON-RPC request that overrides the default ping message reply
- [x] `kb_delayInit` - a JSON-RPC request for scheduling a message text update after some time elapses
- [x] `kb_ping` - a JSON-RPC request that calls the "ping" method to get the stored reply text
- [x] `kb_pong` - a request that triggers an invalid method; always triggers a 500 status response

Incidentally, ports the `@kubelt.platform/ping` service to use `@kubelt/openrpc/component` framework. This was mainly so that I'd have a service to develop without risking the introduction of unintentional breakage.

- [x] add a Durable Object to ping service called `ReplyMessage`
  - this objects stores the message text that is returned in response to a `kb_ping` service request
- [x] add a `kb_init` service method
  - this allows the default ping message reply to be set on the `ReplyMessage` durable object
- [x] add a `kb_delayInit` service method
  - this allows the scheduled update of the ping message reply
  - the message to store is provided as the `message` parameter
  - the time delay before the scheduled update is provided as the `delay` parameter
  - the delay is defined to be an object of the form: `{ years: <n>, ..., seconds: <n> }`
  - the service calls the `schedule` method on the `ReplyMessage` DO to store the pending message and schedule the invocation of the update method (marked using the new `@alarm` decorator)
  - after the delay elapses, the message is updated and subsequent calls to `kb_ping` should return the new message

More incidentals:

- [x] remove `@kubelt/openrpc/middleware` that are not used and unlikely to be developed further:
  - `mwOort`
  - `mwDatadog`
- [x] the starbase application node (durable object) was missing on `main` at the point this branch was created; it is restored here by picking it off of another branch

## testing

After running the `@kubelt.platform/ping` service, try sending these requests:
- `kb_ping`: returns the default ping reply message `hola!`
- `kb_init`: sets the new ping reply message to `pong`
- `kb_ping`: returns the new reply message `pong`
- `kb_delayInit`: schedules a message after to set new message `hasta la vista` after 3 seconds
- `kb_ping`: (before 3 seconds) returns the reply message `pong`
- `kb_ping`: (after 3 seconds) returns the reply message `hasta la vista`

Run and send messages to the service by:
```
$ cd platform/ping
$ npx yarn
$ npx yarn run dev
```
From another console, in the `platform/ping` directory:
```
$ curl http://localhost:XXXX/jsonrpc -d @curl/kb_ping.json
```


packages/openrpc/component/index.ts
```diff
@@ -361,6 +367,11 @@ export function component(schema: Readonly<RpcSchema>) {
             // Gets all state stored in the object.
             //console.log(await this._state.storage.list());
 
+            // If any alarm was set, make sure it's scheduled.
+            if (alarm.timestamp) {
+              this.storage.setAlaram(alarm.timestamp)
```
szkl, line 372: typo

szkl (approved):

PR #1184, opened 2022-12-12
feat(sb): more edges integration, flesh out API methods

crimeminister:
# Description

## @kubelt/platform.starbase

- [x] now depends on `@kubelt/urns` package

### kb_appCreate

- [x] extract ID of application owner from JWT `sub` claim
- [x] required `clientName` parameter when constructing a new application node 
- [x] no longer generates/returns the client secret
  - you must now call `kb_appRotateSecret` to generate the OAuth client secret
- [x] returns the `clientId` for the application
  - this is used in subsequent calls as the app identifier
- [x] calls the `edges` service to create a link between account node and app node

### kb_appDelete

- [x] removes lookup table entry (from client ID to app node ID)
- [x] removes edge linking account node to app node

### kb_appList

- [x] account to list applications for is determined by the JWT `sub` claim

### kb_appRotateSecret

- [x] generates an OAuth client secret
  - returns the plaintext secret _once_ in response to this call
  - secret is stored hashed, and never handled as plaintext in the system again
- [x] adds a separate `secret` field to the application node
  - this field stores the current hashed application secret
  - secret rotation is _destructive_, overwriting the stored secret
- [x] exposes a `rotateSecret` method from the application node
 
## @kubelt/platform.edges

- [x] expect to receive a full edge tag URN when making an edge

## @kubelt/urns

- [x] add `ApplicationURN` type
  - used to name a starbase application node
- [x] add `ApplicationURNSpace`
  - utility for working with application URNs (membership testing, construction)

### DIFF
--- a/platform/starbase/curl/kb_appCreate.json
+++ b/platform/starbase/curl/kb_appCreate.json
@@ -3,6 +3,6 @@
   "id": 1,
   "method": "kb_appCreate",
   "params": {
-    "ownerId": "robert"
+    "clientName": "foobar App"
   }
 }
--- a/packages/openrpc/component/alarm.ts
+++ b/packages/openrpc/component/alarm.ts
@@ -11,7 +11,6 @@ import { add } from 'date-fns'
  * A utility class for setting an alarm time in a durable object.
  */
 export class RpcAlarm {
-
   private _timestamp: Date | undefined
 
   /**
@@ -55,5 +54,4 @@ export class RpcAlarm {
     this._timestamp = add(Date.now(), duration)
     return this._timestamp
   }
-
 }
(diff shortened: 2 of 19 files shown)


platform/starbase/curl/kb_appCreate.json
```diff
@@ -3,6 +3,6 @@
   "id": 1,
   "method": "kb_appCreate",
   "params": {
-    "ownerId": "robert"
+    "clientName": "foobar App"
```
crimeminister, line 6: @maurerbot this call now takes `clientName` parameter.

maurerbot (approved):

PR #1198, opened 2022-12-13
feat(sb): all applications to be published

crimeminister:
# Description

Allow applications to be "published" using the `kb_appPublish` RPC method of `@kubelt/platform.starbase` service.

## @kubelt/platform.starbase

- [x] split out RPC service error definitions into separate `error` module
- [x] add methods on application node
  - [x] `hasSecret` returns true iff application secret is not empty
  - [x] `publish` sets the publication state of application to true or false
- [x] comments out application of `@requiredScope()` decorator in application node until it's better defined

### kb_appCreate

- [x] rename `clientName` parameter to `name` to match already-stored field

### kb_initPlatform

- [x] hash the secret provided in fixture data
- [x] update application node `init` method to store the hashed secret
  - this is only used for the case where loading fixture data, and should be removed once fixture objects are no longer necessary

### kb_appPublish

- [x] update parameters to include `clientId` (to identify app being published)
- [x] update parameters to include `published` flag (to indicate whether app is published or not)
- [x] adds app data validity check before allowing app to be published
  - the app secret must be set
  - the app name must be set

### DIFF
--- a/platform/starbase/curl/kb_appPublish.json
+++ b/platform/starbase/curl/kb_appPublish.json
@@ -1,0 +1,9 @@
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "kb_appPublish",
+  "params": {
+    "clientId": "53ab56010b446e62746309dba44c35c2",
+    "published": true
+  }
+}
--- a/platform/starbase/curl/kb_appRotateSecret.json
+++ b/platform/starbase/curl/kb_appRotateSecret.json
@@ -3,6 +3,6 @@
   "id": 1,
   "method": "kb_appRotateSecret",
   "params": {
-    "clientId": "9300468a8b0d9ae54ee13e5aa0aad650"
+    "clientId": "53ab56010b446e62746309dba44c35c2"
   }
 }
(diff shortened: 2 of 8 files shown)


platform/starbase/curl/kb_appCreate.json
```diff
@@ -3,6 +3,6 @@
   "id": 1,
   "method": "kb_appCreate",
   "params": {
-    "clientName": "foobar App"
+    "name": "foobar App"
```
crimeminister, line 6: @maurerbot I changed this RPC parameter name to match the name of the field that was already being stored, assuming that this is just the "application name". Or does there need to be a separate OAuth name field?

maurerbot: clientName is actual better I think? doesn't need to be seperate.

maurerbot (changes requested):

platform/starbase/curl/kb_appCreate.json
crimeminister: I guess I was just worried about possible confusion; both `clientId` and `clientSecret` fields are described by OAuth flow and have significance in that protocol, but `clientName` is an arbitrary user-supplied name for the app that (AFAIK) isn't relevant to the OAuth flow. I am good to keep the new name, am possibly reading too much into this naming choice.

maurerbot (approved):

PR #1218, opened 2022-12-14
chore(object): remove object worker

szkl:
This worker is unused and there is no functional requirement to maintain it.


szkl:
I don't see a reason to maintain. It is not updated. Is there any plan for general-purpose object storage?

PR #1222, opened 2022-12-14
chore(utils): introduce `@kubelt/utils` package

szkl:
Need to settle https://github.com/kubelt/kubelt/pull/1218 before merging this.


crimeminister:
@szkl Will this package be the right place for platform constants, e.g. header names? Or do you think it makes more sense to keep things like in a more specific package (`@kubelt/http`, `@kubelt/constants`)?

szkl:
> @szkl Will this package be the right place for platform constants, e.g. header names? Or do you think it makes more sense to keep things like in a more specific package (`@kubelt/http`, `@kubelt/constants`)?

I think it is too early now. I don't expect this header to be passed to platform workers because once the access worker verifies the token then the rest of the process doesn't have to pass it around and have it verified over and over again. I just kept the constant definition not to break anything at the moment.

I say let's wait until if there's going to be a practical need.

maurerbot (approved):

maurerbot (approved):

PR #1272, opened 2022-12-20
fix(nftar): NFTar Speedup Attempts

### LINKED ISSUE #1264: chore(nftar): Speed pt1: add CF images to voucher for web and keep ipfs for token
- [ ] store pfp in CF and add cfPfp prop in voucher 
- [ ] store cover in CF and cfCover prop in voucher
- [ ] use cfPfp and cfPFP to generate OF preview image (nftar og image endpoint)

Notes:
- cfPfp and cfCover prop might not be needed if CID can be used in CF images

alfl:
# Description

Speedup attempts for NFTar OG tag generation.

- Closes #1264

## Type of change

# How Has This Been Tested?

Local, goerli

# Checklist:

- [na] I have made corresponding changes to the documentation website
- [na] I have made corresponding changes to the literal docs
- [x] I have added tests that prove my fix is effective or that my feature works


projects/nftar/src/app.js
```diff
@@ -306,116 +308,22 @@ jsonrpc.method('3id_genPFP', async (ctx, next) => {
     console.log(`Crypto took ${t1 - t0} milliseconds.`);
 
     // Generate the response to send to the client.
-    ctx.body = {
-      metadata: metadata.data,
-      voucher,
-      signature,
+    const body = {
+        metadata: metadata.data,
+        voucher,
```
maurerbot, line 313: @alfl if voucher has the cf images as props too we can use that in the bffs

maurerbot (commented):
dirty solution is now just add cf props to the voucher and we make the associated changes in bff to use that as profile image and cover while bootstrapping an account

alfl:
> dirty solution is now just add cf props to the voucher and we make the associated changes in bff to use that as profile image and cover while bootstrapping an account
> 
> Let's land this tomorrow please

This doesn't work. I have it implemented locally -- CF Images has bad performance issues for a process like this. Looking at R2.

alfl:
> hm. it's slow to finalize the url? no fire and forget model?

IPFS is 20-30 seconds, CF Images is 10-15 seconds. The cover and PFP uploads have to finish before OG generation. So fire-and-forget hides the race condition but then you have to make sure you wait until they're done before doing OG.

Makes more sense IMO to just generate the OG when you know the uploads are done, but now you're baking in these delays.

Quick Google shows performance is a known issue with CF Images -- it's for simple web use cases.

R2 might be faster and give us some advantages in serving the OGs straight from the public public.

alfl:
> if doing r2 create a "content" worker in /services ...

web3 CMS, yes 😂 

There was a feature request about image sharing on Farcaster -- might be able to help dwr out :)

alfl:
Still doing some smoke tests but speeds much improved:

- Moved image storage to R2, but preserved the `ipfs://` URLs embedded in the NFT metadata
- Parallelized some requests and generally made things async fire-and-forget calls.
- Reducing PFP image size gave us ~3x-5x speedup, reduced network traffic delays, and container memory consumption.

There is more juice to squeeze in moving R2 to a custom domain, which lifts some of Cloudflare's rate limits.

alfl:
Also updated URLs and removed invite code paths. More cleanup possible too.

szkl (approved):

PR #1275, opened 2022-12-20
feat(console): rotate secret

Cosmin-Parvulescu:
# Description

Added functionality for displaying application secrets on application creation.

## Type of change

# How Has This Been Tested?

Flow

# Checklist:


platform/starbase/src/index.ts
```diff
@@ -893,6 +895,59 @@ const kb_appPublish = openrpc.method(schema, {
...
+        return openrpc.error(request, detail)
+      }
+
+      const app = await openrpc.discover(sbApplication, {
+        id: appId,
+        tag: 'starbase-app',
+      })
+      invariant(appId === app.$.id, 'object IDs must match')
+
+      // NB: this returns an empty object if the application is not
```
crimeminister, line 936: This comment is not accurate when calling `fetch()` rather than `profile()`.

Cosmin-Parvulescu: Addressed :)

apps/console/app/routes/apps/$clientId/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+      title: string
+    }
+  }
+
+  let rotatedSecret
+  if (!appDetails.secretTimestamp) {
+    rotatedSecret = await starbaseClient.kb_appRotateSecret(appDetails.clientId)
+
+    // For some reason secret is
+    // secret:{actualSecret}
```
crimeminister, line 40: The prefix is there just as an aide to users; when they're moving these values (client ID, client secret), the prefix should help distinguish between them, rather then the user having to distinguish between them by e.g. length.

Cosmin-Parvulescu: This should only be displayed once for the user, so it's probably OK to handle the case on client side. I guess the fail case is if it ever changes the client side split might fail and error out, but I'm a firm believer in letting errors explode to indicate a problem so that's OK with me.

Cosmin-Parvulescu: I also updated the comment with your explanation as it makes more sense than `for some reason` :)

crimeminister (approved):

PR #1283, opened 2022-12-20
feat(address): use edges to store and retrieve account links

crimeminister:
# Description

Uses the `@kubelt/platform.edges` service to link and unlink address and account nodes.

## @kubelt/types

- [x] add barebones platform types package
  - initial type added that was troublesome to import from service package where currently defined
  - this means there's a duplication type definition for the platform; we should discuss how to resolve it

## @kubelt/graph

- [x] consolidate all edge types in `@kubelt/graph/edges`
- [x] add linking utilities for specific edge types in `@kubelt/graph/link`
  - `linkAccountAddress()`, `unlinkAccountAddress()` - create edges linking account and address node
  - `linkAccountApp()`, `unlinkAccountApp()` - create edges linking account and starbase application node
  - `listAddresses()` - list addresses associated with an account (incl. filtering on address type)
  - these utilities are now the preferred way to interact with the `edges` service
- [x] rework `edges()` call to accept an `EdgeQuery`
  - allows for better filtering of the returned edge set

Edge queries now accept an `EdgeQuery` having optional properties:
- `id`: the "base" URN of a node (without f-, q-, r-components)
- `tag`: the type tag of the edges to return
- `dir`: the direction of the edge (is the `id` for the "source" or "destination" node?)
- `src`: a filter on the properties of the source node URN
- `dst`: a filter on the properties of the destination node URN

## @kubelt/platform.address

- [x] updates `kb_setAccount` method to create a link between address and account
- [x] updates `kb_unsetAccount` method to remove a link between address and account
- [x] add placeholder `getWebHook` and `setWebHook` methods in `crypto` node
  - incidentally, changed case of "webhook" in schema to `WebHook` everywhere for consistency

## @kubelt/platform.account

- [x] `kb_getAddresses` - get the list of addresses associated with an account
- [ ] `kb_hasAddresses` - checks whether or not an account "owns" certain addresses
  - currently just scaffolding, to be implemented in a subsequent PR

## @kubelt/platform.starbase

- [x] update call to `graph.edges()` to use `EdgeQuery`

## docs/graph

- [x] add placeholder section for graph-related documentation


platform/address/src/jsonrpc/methods/getAccountList.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export default async (
+  service: Readonly<RpcService>,
+  request: Readonly<RpcRequest>,
+  context: Readonly<RpcContext>
+) => {
+  const edges: Fetcher = context.get('Edges')
+  const nodeClient = context.get('node_client')
+
+  // Get the address associated with the X-3RN included in the request.
```
crimeminister, line 17: @szkl I assume I should be dealing with the address specified in the request, which afaict is available on the context as the `name` value. Does this seem right?

platform/address/src/schemas/crypto.ts
```diff
@@ -208,32 +208,31 @@ export const schema: RpcSchema = {
...
-    //     {
-    //       $ref: '#/components/contentDescriptors/Webhook',
-    //     },
-    //   ],
-    //   result: {
-    //     name: 'set webhook result',
-    //     schema: {},
-    //   },
-    // },
+    {
```
maurerbot, line 211: can we comment this back out?

maurerbot (changes requested):
Missing account methods to get addresses

platform/address/src/schemas/crypto.ts
crimeminister: Sure, I added placeholder methods in the object to match, I'll comment those out too.

crimeminister:
> 

I added the `kb_getAccountList` method to get the list of accounts for an address. Is there another way you'd like to be able to get that information?

crimeminister:
If you're okay with this PR, I can add a search-based means of retrieving accounts for a given address; currently the assumption is that the address is determined by `X-3RN` header, but perhaps we want a more general facility than that?

platform/address/src/jsonrpc/middlewares/parse3RN.ts
```diff
@@ -21,6 +21,7 @@ export default (request: Readonly<Request>, context: RpcContext) => {
 
   const { rcomponent, qcomponent } = AddressURNSpace.parse(urn)
   const rparams = new URLSearchParams(rcomponent || '')
+  context.set('rparams', rparams)
```
crimeminister, line 24: @szkl I just wanted to call out this change to the `parse3RN` middleware. I will use the r-components in the RPC method handler to construct a URN for the address that has these values appended so they can be indexed in the database.

docs/CONTENTS.md
```diff
@@ -12,6 +12,7 @@
   * [Console Reference](3id/console/reference.md)
 * [Passport](3id/passport/index.md)
 * [Galaxy](3id/galaxy/index.md)
+* [Graph](3id/graph/index.md)
```
alfl, line 15: `reference.md` link needed one below to ACTIVATE! MERMAID! 🧜‍♀️ 

```
    * [Graph Reference](3id/graph/reference.md)
```

crimeminister: Added!

packages/graph/util.ts
```diff
@@ -0,0 +1,168 @@
+// @kubelt/graph:util.ts
+
+/**
+ * Platform utilities for creating, removing, and listing various kinds
+ * of platform links without worrying about the details of making calls
+ * to the "edges" service.
+ */
+
+// TODO This enum is exported from @kubelt/platform.address, but can't
```
crimeminister, line 9: @szkl I'd like to use the `CryptoAddressType` enum from `@kubelt/platform.address`. I can go ahead and figure out what needs to be done for that to work (the first attempt resulted in an error), but before I do I thought I should ask what you thought about a `@kubelt/types` package for types that are common to the platform as a whole? Or maybe there's another package where these types might fit?

crimeminister: I made `@kubelt/types` for this purpose.

platform/account/src/jsonrpc/methods/getAddresses.ts
```diff
@@ -0,0 +1,79 @@
...
+  const edgeResult = await listAddresses(edges, account, type)
+
+  if (Array.isArray(edgeResult)) {
+    // The source nodes in the returned edges are the URNs of the
+    // account nodes.
+    const addresses = edgeResult.map((edge: Edge) => {
+      return edge.dst.urn
+    })
+    return openrpc.response(request, {
+      account,
```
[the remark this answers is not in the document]
crimeminister: I have the habit of trying to make responses "self-contained", minimizing the amount of contextual information required to interpret the result data. In my experience it isn't always the case the assumptions about the calling and receiving environments are borne out, especially as code is modified over time. That said, if we don't need or want the extra properties in the result, I am happy to remove them.

platform/edges/src/db/impl/insert.ts
```diff
@@ -95,18 +98,26 @@ export async function node(g: Graph, urn: AnyURN): Promise<NodeRecord> {
       selects.push(qcStmt.bind(key, value))
     }
     const qcResult = await g.db.batch(selects)
-    // Collect an array of row IDs for the table rows containing q-components.
-    const rowIds = _.map(qcResult[0].results, (o) => {
-      return _.get(o, 'id', -1)
-    })
+    // Collect an array of row IDs from the table rows containing q-components.
+    const allRowIds = _.reduce(qcResult, (acc: number[], o: Record<string, unknown>): number[] => {
```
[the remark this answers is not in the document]
crimeminister: Lots of JS scars, mainly. I have been using utility libraries to paper over weaknesses in the JS "standard library" for so long I reflexively reach for those tools. I have noticed some problems with using lodash as far as type analysis goes (e.g. `_.isUndefined()` and `x === undefined` seem not to be equivalent for type analysis purposes) so I am stripping out usage of lodash as I encounter it.

platform/account/src/jsonrpc/methods/getAddresses.ts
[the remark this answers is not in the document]
crimeminister: Updated.

packages/platform-middleware/inputValidators.ts
```diff
@@ -15,3 +16,19 @@ export const AccountURNInput = z.custom<AccountURN>((input) => {
   }
   return input as AccountURN
 })
+
+export const CryptoAddressTypeInput = z.custom<CryptoAddressType>((input) => {
```
crimeminister, line 20: @szkl Added this input validator; was hoping to get your 👀 until I'm more familiar with `trpc`.

platform/account/src/jsonrpc/methods/getAddresses.ts
```diff
@@ -0,0 +1,49 @@
...
+  account: inputValidators.AccountURNInput,
+  type: inputValidators.CryptoAddressTypeInput.optional(),
+})
+
+export const AddressList = z.array(inputValidators.AddressURNInput)
+
+// TODO move to types package
+export type AddressList = z.infer<typeof AddressList>
+
+export const getAddressesMethod = async ({
```
crimeminister, line 26: @szkl I'd also appreciate if you could 👀 this RPC method implementation.

maurerbot (approved):

PR #1304, opened 2022-12-22
feat(console): API key rolling implementation

### LINKED ISSUE #1190: feat(galaxy): API key
# Why

Galaxy needs an authentication method for customers to make API call to Galaxy

# What

An API Key would be given for each app created in console. This key should:
- allow access to the authenticated endpoints

# How

- [ ] Generate an API Key for new starbase app. API key will be a JWT, encoding the app account id in the `sub`; the signing keypair will be generated/stored in starbase. 
- [ ] Ability to rotate api key
- [ ] Middleware in galaxy for authenticating API keys, validating JWT signature in starbase through new validation endpoint.

# Notes

https://www.figma.com/file/EqUEbCGHGnZXDMSUbrxhX7/Kubelt?node-id=5283%3A14983&t=jlSGrVGYrwxKz42g-0

betimshahini:
# Description

Implements user-facing feature of being able to roll the API key. Also includes refactorings to remove need for worker method to clear the secrets to make it work in Remix.

- Closes #1190 

## Type of change

# How Has This Been Tested?

- [x] Create new app -> Both secrets show up and date is set to current
- [x] Load app -> No secrets show up but timestamps reflect last rolled datetime
- [x] Roll API key -> New API key shows up and its timestamp gets updated
- [x] Roll Client Secret -> New Client Secret shows up and its timestamp gets updated

# Checklist:

### DIFF
--- a/apps/console/app/routes/apps/$clientId/index.tsx
+++ b/apps/console/app/routes/apps/$clientId/index.tsx
@@ -3,9 +3,10 @@
  */
 
 import { ActionFunction, json, LoaderFunction } from '@remix-run/cloudflare'
-import { useLoaderData, useSubmit } from '@remix-run/react'
+import { useActionData, useLoaderData, useSubmit } from '@remix-run/react'
+import invariant from 'tiny-invariant'
 import { ApplicationDashboard } from '~/components/Applications/Dashboard/ApplicationDashboard'
-import { getStarbaseClient } from '~/utilities/platform.server'
+import { getStarbaseClient, StarbaseClient } from '~/utilities/platform.server'
 import { requireJWT } from '~/utilities/session.server'
 
 // Component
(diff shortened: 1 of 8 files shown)


apps/console/app/routes/apps/$clientId/index.tsx
```diff
@@ -67,44 +58,74 @@ export const action: ActionFunction = async ({ request, params }) => {
...
-  switch (op) {
-    case 'roll_api_key':
-      break
-    case 'roll_app_secret':
-      await starbaseClient.kb_appClearSecret(params.clientId)
-      break
+const RollType = {
+  RollAPIKey: 'roll_api_key',
+  RollClientSecret: 'roll_app_secret',
+  RollBothSecrets: 'roll_both'
```
maurerbot, line 72: when would you roll both?

platform/starbase/src/nodes/application/index.ts
```diff
@@ -320,13 +306,15 @@ export class StarbaseApplication {
   //@requiredScope()
   @requiredField('apiKey', [FieldAccess.Read, FieldAccess.Write])
   @requiredField('apiKeySigningKeyPair', [FieldAccess.Read, FieldAccess.Write])
+  @requiredField('apiKeyTimestamp', [FieldAccess.Write])
```
maurerbot, line 309: why not store with the key?

maurerbot: if you are always setting and getting them together just store as an object.

@crimeminister at this point I think we should just drop the `@requiredField` decorator. We don't have any plans for field level scopes and the DO native storage API is good.

Cosmin-Parvulescu (changes requested):
Looks good; requesting change because there's some whitespace that will fail the build, so maybe `prettier` needs to be run on the files. Glad that extra clearing got taken out. 💪

platform/starbase/src/nodes/application/index.ts
Cosmin-Parvulescu: > @maurerbot @crimeminister can a field (in the context of our framework) store an object? I thought we couldn't do `output.set(somefield, some_convoluted_object)`.

It can store, check out `app` field in the same DO

Cosmin-Parvulescu:
> > Looks good; requesting change because there's some whitespace that will fail the build, so maybe `prettier` needs to be run on the files. Glad that extra clearing got taken out. 💪
> 
> @Cosmin-Parvulescu the PR builds run a format check and they don't look like they've failed. What build are you referring to?

Hmm, might be remnants of an older flow where no newlines at end of file would fail on merge to main.

Cosmin-Parvulescu (approved):

PR #1308, opened 2022-12-23
feat(passport): wip oauth/google

szkl:


apps/passport/.dev.vars.example
```diff
@@ -1,3 +1,7 @@
 THREEID_APP_URL="http://localhost:9797"
 APIKEY_ALCHEMY_PUBLIC=""
 SECRET_SESSION_SALT="foobar"
+
+GOOGLE_OAUTH_CLIENT_ID=""
+GOOGLE_OAUTH_CLIENT_SECRET=""
```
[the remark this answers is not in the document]
szkl: Just fixed.

apps/passport/app/auth.server.ts
```diff
@@ -0,0 +1,38 @@
+import { createCookieSessionStorage } from "@remix-run/cloudflare";
+import { Authenticator } from 'remix-auth'
+import { GoogleStrategy } from 'remix-auth-google'
+
+const sessionSecret = SECRET_SESSION_SALT
+if (!sessionSecret) {
+  throw new Error('SECRET_SESSION_SALT must be set')
+}
+
+export const sessionStorage = createCookieSessionStorage({
```
maurerbot, line 10: seems like we need to name these per authentication methods or it should be just part of the google route?

can't we import the session server cookie here?

apps/passport/app/components/authentication/Authentication.tsx
```diff
@@ -43,16 +45,11 @@ export function Authentication({
...
-        btnSize={'xxl'}
-        disabled
-      >
-        @ Connect with Email
-        <br />
-        (coming soon)
-      </Button> */}
+      <Form action='/authenticate/google' method='post'>
+        <Button className={'button'} btnType={'secondary-alt'} btnSize={'xxl'} isSubmit={true}>
+          Connect With Google
```
maurerbot, line 50: neeed a google button component and include the google logo as per design?

apps/passport/app/routes/authenticate/google.tsx
```diff
@@ -0,0 +1,6 @@
+import type { ActionArgs, ActionFunction } from '@remix-run/cloudflare'
+import { authenticator } from '~/auth.server'
+
+export const action: ActionFunction = ({ request }: ActionArgs) => {
```
maurerbot, line 4: if this is a loader then the button doesn't need a form?

platform/address/src/jsonrpc/methods/setData.ts
```diff
@@ -0,0 +1,15 @@
...
+import { SetDataParams } from '../../types'
+
+export default async (
+  service: Readonly<RpcService>,
+  request: Readonly<RpcRequest>,
+  context: Readonly<RpcContext>
+) => {
+  const nodeClient = context.get('node_client')
+  const [data] = request.params as SetDataParams
+  await nodeClient.setData({ data })
```
maurerbot, line 13: feels dangerous. could override other specific fields? What is being stored in this case?

maurerbot (changes requested):

apps/passport/app/auth.server.ts
szkl: > seems like we need to name these per authentication methods or it should be just part of the google route?

I don't think that's needed because a single strategy will be used at a time and the cookie is only needed just because the `remix-auth` requires it and the access tokens from third-party sites are not used for login state so they're not really useful as a cookie on the client.

> can't we import the session server cookie here?

That's `3ID_SESSION` if I understood correctly. I don't think OAuth should use the same cookie as it is not used for logged-in state representation.

apps/passport/app/components/authentication/Authentication.tsx
szkl: Yes. Will do that.

apps/passport/app/routes/authenticate/google.tsx
szkl: Yes. The form is only used for navigating the browser to the route.

platform/address/src/jsonrpc/methods/setData.ts
szkl:
```json
{
  "accessToken": "ACCESS TOKEN",
  "profile": {
    "provider": "google",
    "id": "116638464152193423777",
    "displayName": "Sönmez Kartal",
    "name": {
      "familyName": "Kartal",
      "givenName": "Sönmez"
    },
    "emails": [
      {
        "value": "szkl89@gmail.com"
      }
    ],
    "photos": [
      {
        "value": "https://lh3.googleusercontent.com/a/AEdFTp7sDkuoq1MR2rNcvLVlJE6tKQ4ouEC9cGVmXZ2jqyc=s96-c"
      }
    ],
    "_json": {
      "sub": "116638464152193423777",
      "name": "Sönmez Kartal",
      "given_name": "Sönmez",
      "family_name": "Kartal",
      "picture": "https://lh3.googleusercontent.com/a/AEdFTp7sDkuoq1MR2rNcvLVlJE6tKQ4ouEC9cGVmXZ2jqyc=s96-c",
      "email": "szkl89@gmail.com",
      "email_verified": true,
      "locale": "en"
    }
  },
  "extraParams": {
    "expires_in": 3599,
    "scope": "openid https://www.googleapis.com/auth/userinfo.profile https://www.googleapis.com/auth/userinfo.email",
    "token_type": "Bearer",
    "id_token": "some JWT"
  }
}
```

szkl: The data is the payload from the OAuth server for the `OAuthAddress` node. It writes to the `data` field only in the node storage.

apps/passport/app/components/authentication/Authentication.tsx
szkl: Fixed. I left out the storybook script for now.

apps/passport/app/routes/authenticate/google.tsx
szkl: I think an action and a form submission fit better for a button semantically but it is pretty much similar to an anchor linking to the route but that's more like visiting by clicking on a link. What to do?

szkl:
image

apps/passport/app/auth.server.ts
maurerbot: so other oauth providers will use the same `GoogleStrategy`?

apps/passport/app/routes/authenticate/google.tsx
maurerbot: I'm indifferent at this point so a form will work

platform/address/src/jsonrpc/methods/setData.ts
maurerbot: ah so `setData` is really `setOauthData`? Maybe we should be more specific?

szkl: That's what I proposed the other day. This method is not just meant for `OAuth` data. It can be used to set third-party data in other node types. I'll just rename it as it seems like just confusing right now.

apps/passport/app/auth.server.ts
szkl: No. They would need a strategy instance of their own type implementing OAuth client for that provider.

maurerbot: that's what I mean. so shouldn't this file be called google?

szkl: Oh. Well, I think this file can be used for all of them with different export names.

alfl (dismissed):

apps/passport/app/routes/authenticate/google/callback.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export const loader: LoaderFunction = async ({ request }: LoaderArgs) => {
+  const {
+    accessToken,
+    refreshToken,
+    extraParams,
+    profile,
+  } = await authenticator.authenticate('google', request) as AuthenticationResult
+
+  const address = AddressURNSpace.urn(profile._json.email) + '?+node_type=oauth?=addr_type=google' as AddressURN
```
maurerbot, line 31: urn should be based on hash of email?

maurerbot (changes requested):

szkl:
The error has been resolved. Good for merge now.

maurerbot (approved):

PR #1323, opened 2022-12-29
chore(account): define durable object state for `IttyDurable`

szkl:


szkl (commented):
`IttyDurableObjectState` can be moved to a common package, of course. This branch is just to get it working.

Issue #1114, opened 2022-12-02
feat(threeid): connected crypto account settings tab


maurerbot:
# Why

Users need a section to manage their connected accounts

# What

In the settings page we need to add a connected accounts tab that loads all connected addresses for the logged in user account. 

# How

- [ ] Connected accounts tab in settings
- [ ] Address<=->Account edges
- [ ] Add / remove accounts

# Notes

- https://www.figma.com/file/EqUEbCGHGnZXDMSUbrxhX7/Kubelt?node-id=4727%3A15130&t=qtG0ujtrZxMvsa81-0
- depends on #1113

maurerbot:
Related to #1115

PR #1327, opened 2022-12-29
feat(account): add method to check if list of addresses is owned by an account

crimeminister:
# Description

Adds a method to `@kubelt/platform.account` to check whether or not some addresses belong to an account.

## @kubelt/platform.account

- [x] remove now-unneeded JSON-RPC error defs
- [x] extract `AddressList` zod schema and type export to `middlewares/addressList`
  - updates `getAddresses` to use now-shared schema / type
- [x]  add `hasAddresses` RPC method that:
  - accepts an `account` and a list of `address` URNs
  - fetches the list of addresses owned by the account from `edges` service
  - returns a flag indicating whether or not input addresses are all owned by the account

Example request body:
```json
{
  "account": "urn:threeid:account/foo",
  "addresses": [
    "urn:threeid:address/0xb794f5ea0ba39494ce839613fffba74279579268"
  ]
}
```
Example cURL against the `account` service:
```shell
$ curl -H "Content-Type: application/json" http://localhost:9494/trpc/hasAddresses -d @curl/hasAddresses.json
```

NB: you will first need to create the linkage between an account and an address by invoking the `kb_setAccount` method on the `address` service for this to work:
```shell
$ curl http://locahost:9191/jsonrpc -d @$root/platform/address/curl/kb_setAccount.json -H "X-3RN: urn:threeid:address/0x...?+node_type=crypto&addr_type=ethereum"
```


crimeminister:
@poolsar42 does this API fit your needs re checking whether or not some addresses are owned by an account?

poolsar42:
> @poolsar42 does this API fit your needs re checking whether or not some addresses are owned by an account?

from #1114 I see that I need get list of accounts from any address if I understand correctly

Here is visa-versa?

poolsar42:
Nevermind, I need list of addresses. This looks correct then

poolsar42 (approved):

platform/account/src/jsonrpc/methods/hasAddresses.ts
```diff
@@ -0,0 +1,65 @@
...
+import { Context } from '../../context'
+import { AddressURNSpace } from '@kubelt/urns/address'
+
+import type { AddressList } from '../middlewares/addressList'
+import type { Edge } from '@kubelt/graph'
+import type { AccountURN } from '@kubelt/urns/account'
+import type { AddressURN } from '@kubelt/urns/address'
+
+// Should this live in @kubelt/platform-middlewares/inputValidators?
+export const AddressListInput = z.custom<AddressList>((input) => {
```
maurerbot, line 14: I think we have a validator that does this? if not we should co-lo this one with those?

platform/account/src/jsonrpc/methods/hasAddresses.ts
```diff
@@ -0,0 +1,65 @@
...
+  input,
+  ctx,
+}: {
+  input: HasAddressesParams
+  ctx: Context
+}) => {
+  // Return the list of edges between the account node and any address
+  // nodes. Don't filter the addresses by type, we want them all (the
+  // total number is normally going to be small).
+  const edgeList = await listAddresses(ctx.Edges, input.account)
```
maurerbot, line 46: how is `ctx.Edges` defined?

maurerbot (changes requested):
LGTM. just a couple questions

platform/account/src/jsonrpc/methods/hasAddresses.ts
```diff
@@ -0,0 +1,65 @@
...
+import type { AccountURN } from '@kubelt/urns/account'
+import type { AddressURN } from '@kubelt/urns/address'
+
+// Should this live in @kubelt/platform-middlewares/inputValidators?
+export const AddressListInput = z.custom<AddressList>((input) => {
+  if (!Array.isArray(input)) {
+    throw new Error('address list must be an array')
+  }
+  input.forEach((address) => {
+    if (AddressURNSpace.parse(address as AddressURN) === null) {
```
szkl, line 19: `AddressURNSpace.is()` is a better fit.

platform/account/src/jsonrpc/methods/hasAddresses.ts
crimeminister: We have a validator for a single AddressURN, while this is check that an array contains all valid addresses. I don't know that it will be useful outside of this specific method, so kept it local. Don't mind moving it to `platform-middleware` if that makes sense?

platform/account/src/jsonrpc/methods/hasAddresses.ts
crimeminister: I copied what was done for `AddressURNInput` validator; I will change it however.

platform/account/src/jsonrpc/methods/hasAddresses.ts
crimeminister: It's defined as part of the context [here](https://github.com/kubelt/kubelt/blob/main/platform/account/src/context.ts#L14).

maurerbot (approved):

platform/account/package.json
```diff
@@ -38,6 +38,9 @@
     "@kubelt/types": "workspace:*",
     "@trpc/server": "10.7.0",
     "do-proxy": "1.3.1",
+    "itty-durable": "1.6.0",
+    "itty-router-extras": "0.4.2",
```
crimeminister, line 42: @maurerbot do we still want/need `itty-router-extras`?

maurerbot: no we can drop these

PR #1330, opened 2022-12-30
feat(access): create edges to link sessions

crimeminister:
# Description

## @kubelt/platform.access

- [x] create an edge linking `Account` and `Access` nodes upon creation of user session
  - the edge is created after token exchange is performed
- [x] add `getSession` service  method to fetch session details
- [x] add `revokeSession` service method to revoke a session
- [x] add `status` node method to return session description
- [x] add `revoke` node method
  - removes all DO state allowing it to be garbage collected
- [x] store extra fields on `Access` node:
  - `creationTime`: the creation time of the session
  - `expiryTime`: the expiry time of the session
  - `expired`: a flag indicating if session has expired
- [x] register `Access` DO alarm on JWT expiry + delta to flag the session as being expired
  - we don't delete state as the object is preserved to be able to show the session to a user
  - sessions are only deleted when `revoke()` is called
  - when a session token is refreshed any scheduled alarm is replaced with new expiry time
- [x] remove defunct openrpc schemas

## @kubelt/platform.account

- [x] add `getSessions` RPC call to list sessions for an account

## @kubelt/graph

- [x] add edge type `EDGE_ACCESS`
- [x] remove utility `linkAccountAddress`
- [x] remove utiliyt `unlinkAccountAddress`
- [x] remove utility `listAddresses`

## notes

- uses the `AccessURN` as the destination "node type" of edges linking an `Account` and `Access` node
  - the `Access` node is our user session storage object
- the original plan for deleting the `account ↔ access` edge when a session expired was to call the edges service directly via a service binding from the DO alarm handler; service bindings do not appear in the DO environment, unlike other binding types, so this proved infeasible
  - using a Worker Queue to allow "edge remove operations" to be scheduled also had issues
  - eventually the API was reshaped to bypass the issue


crimeminister:
@maurerbot In `@kubelt/platform.address` method `setAddress` we require an `X-3RN` header, but in the input parameters we require an explicit `address: AddressURN`. Can we remove one or the other?

crimeminister:
@maurerbot When calling `@kubelt/platform.address` method `setAccount` I am seeing the error:
```
addressNode.class.getAddress is not a function
```
Can you confirm this is happening for you also?

packages/graph/edges.ts
```diff
@@ -22,9 +22,15 @@ export const EDGE_APPLICATION: EdgeURN = edge('owns/app')
  */
 export const EDGE_ADDRESS: EdgeURN = edge('owns/address')
 
+/**
+ * An edge linking an account node (representing a user account) and a
+ * client session node for that account.
+ */
+export const EDGE_ACCESS: EdgeURN = edge('owns/access')
```
maurerbot, line 29: this should now be in access/constants.ts

platform/access/src/jsonrpc/middleware/validators.ts
```diff
@@ -17,3 +21,13 @@ export const tokenValidator = z.custom<{ iss: string; token: string }>(
     }
   }
 )
+
```
maurerbot, line 24: I wonder if this can go to shared-middleware package?

platform/account/src/jsonrpc/methods/getSessions.ts
```diff
@@ -0,0 +1,73 @@
...
+export const GetSessionsMethodInput = z.object({
+  account: inputValidators.AccountURNInput,
+})
+
+export type GetSessionsParams = z.infer<typeof GetSessionsMethodInput>
+
+// Output
+// -----------------------------------------------------------------------------
+
+// Consider moving to inputValidators if potentially reusable.
```
maurerbot, line 27: is this repeated?

maurerbot (changes requested):

crimeminister:
> 

I was thinking here specifically of the `setAddress` method.

platform/access/src/jsonrpc/middleware/validators.ts
crimeminister: Moved!

platform/account/src/jsonrpc/methods/getSessions.ts
crimeminister: DRYed up!

platform/access/src/jsonrpc/methods/revokeSession.ts
```diff
@@ -0,0 +1,52 @@
...
+  input,
+  ctx,
+}: {
+  input: RevokeSessionParams
+  ctx: Context
+}) => {
+  // The ValidateJWT middleware extracts the account URN from the JWT
+  // and places it on the context.
+  const account = ctx?.accountURN
+  if (account === undefined) {
```
maurerbot, line 30: we don't need this check. middleware should throw?

platform/access/src/jsonrpc/methods/exchangeToken.ts
```diff
@@ -79,43 +83,19 @@ export const exchangeTokenMethod = async ({
       scope: [], //scope,
     })
 
+    // Create an edge between Account and Access nodes to record the
+    // existence of a user "session".
+    const edgesClient = createEdgesClient(ctx.Edges)
```
maurerbot, line 88: we have middleware for this in platform-middleware

maurerbot (changes requested):

platform/access/src/jsonrpc/methods/exchangeToken.ts
crimeminister: I borrowed this pattern from elsewhere in the codebase, did you want this updated everywhere it's happening?

platform/access/src/jsonrpc/methods/revokeSession.ts
```diff
@@ -43,7 +40,11 @@ export const revokeSessionMethod = async ({
   // node.
   const edgesClient = createEdgesClient(ctx.Edges)
```
maurerbot, line 41: we have middleware to create edges client too

platform/access/src/jsonrpc/methods/revokeSession.ts
crimeminister: Added `RequireAccount` middleware to enforce the requirement that an account be supplied. This still requires that we use `!` on the `ctx.accountURN` in the method handler though.

platform/access/src/jsonrpc/methods/exchangeToken.ts
crimeminister: Service client is now injected using `InjectEdges` middleware.

platform/access/src/jsonrpc/methods/revokeSession.ts
crimeminister: Service client is now injected using InjectEdges middleware.

platform/access/src/jsonrpc/methods/getSession.ts
```diff
@@ -0,0 +1,37 @@
...
+export const getSessionMethod = async ({
+  input,
+  ctx,
+}: {
+  input: GetSessionParams
+  ctx: Context
+}) => {
+  const name = AccessURNSpace.decode(input)
+
+  const accessNode = await initAccessNodeByName(name, ctx.Access)
```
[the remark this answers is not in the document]
crimeminister: Right now the access node URN is being provided as an input parameter in the `getSession` method and in the `revokeSession` method. Presumably if we initialize the access node client in a middleware we'd want to provide the URN as a header instead, the way the `X-3RN` is used to provide the `addressURN`  in the `address` service. Is that your preference, or does it make sense to pass along the `accessURN` value as an input parameter, and use a utility fn to construct the client?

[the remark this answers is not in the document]
crimeminister: Cf. `InjectAccessNode` middleware.

packages/platform-middleware/jwt.ts
```diff
@@ -27,12 +28,40 @@ export const ValidateJWT: BaseMiddlewareFunction<{
...
+ * is needed if handler must branch on the presence or absence of the
+ * account).
+ *
+ * This middleware throws if the account is not defined on the context
+ * or is not valid.
+ */
+export const RequireAccount: BaseMiddlewareFunction<{
+  accountURN?: string
+}> = ({ ctx, next }) => {
+  if (ctx?.accountURN === undefined) {
```
szkl, line 56: would miss `null` value

platform/access/src/context.ts
```diff
@@ -13,10 +16,16 @@ interface CreateInnerContextOptions
...
   Access: DurableObjectNamespace
   access?: Access
+  accessNode?: DurableObjectStubProxy<Access>
   Authorization: DurableObjectNamespace
   authorization?: Authorization
   Starbase: Fetcher
   starbaseClient?: ReturnType<typeof createStarbaseClient>
+  EDGES: Fetcher
+  // Added by InjectEdges middleware.
+  edgesClient?: ReturnType<typeof createEdgesClient>
```
szkl, line 26: Why optional? If this can be avoided then non-null assertion (!) would be unnecessary in the usage references.

platform/access/src/jsonrpc/middleware/injectAccessNode.ts
```diff
@@ -0,0 +1,35 @@
...
+  const accessNode = await initAccessNodeByName(jwt.iss, ctx.Access)
+
+  if (accessNode === undefined) {
+    throw new Error('unable to get access node client')
+  }
+
+  return await next({
+    ctx: {
+      accessNode,
+      ...ctx,
```
szkl, line 32: The object spread operator usage should come before the overrides.

platform/access/src/jsonrpc/middleware/injectAccessNode.ts
```diff
@@ -0,0 +1,35 @@
+import * as jose from 'jose'
+import { BaseMiddlewareFunction } from '@kubelt/platform-middleware/types'
+import { AccessURNSpace } from '@kubelt/urns/access'
+import { initAccessNodeByName } from '../../nodes'
+import { Context } from '../../context'
+
+export const InjectAccessNode: BaseMiddlewareFunction<Context> = async ({
```
szkl, line 7: The other workers used a setter name pattern. E.g. `setAccessNode`.

platform/access/src/nodes/access.ts
```diff
@@ -54,9 +59,30 @@ export default class Authorization extends DOProxy {
...
+    await this.state.storage.deleteAll()
+    return true
+  }
+
+  async status(): Promise<SessionDetails> {
+    const expired = await this.state.storage.get<boolean>('expired')
+    const expiryTime = await this.state.storage.get<number>('expiryTime')
+    let expiry
+    if (expiryTime) {
+      expiry = new Date(expiryTime).toUTCString()
```
szkl, line 76: Wouldn't it be better to have a UNIX timestamp instead?

platform/access/src/nodes/access.ts
```diff
@@ -71,12 +97,31 @@ const generate = async (
     storage.put('scope', scope),
   ])
 
+  // We only set the creation time once, not when refreshing the token /
+  // session.
+  const creationTime = storage.get('creationTime')
+  if (creationTime === undefined) {
```
szkl, line 103: Strict equality (===) would miss `null` value, also any other empty value of other types.

Usually `!creationTime` expression is good enough to cover all nullish conditions and shorter than `creationTime == null`.

platform/access/wrangler.toml
```diff
@@ -9,6 +9,7 @@ durable_objects.bindings = [
   { name = "Authorization", class_name = "Authorization" },
 ]
 services = [
+  { binding = "EDGES", service = "edges" },
```
szkl, line 12: `Edges`

szkl (changes requested):
It is not JSON-RPC anymore. I think the request body snippets in `curl` directory are useless now.

crimeminister:
> It is not JSON-RPC anymore. I think the request body snippets in `curl` directory are useless now.

I'll remove them. Do we have any other examples in docs or files to demonstrate usage that we can replace them with?

platform/access/wrangler.toml
crimeminister: This has to match the binding name defined in the [platform middleware](https://github.com/kubelt/kubelt/blob/main/packages/platform-middleware/edges.ts#L5) or type errors occur.

platform/access/src/nodes/access.ts
crimeminister: I try to avoid passing around integer timestamps outside of a service because they come with no context attached and are harder to interpret and debug, i.e. are we dealing with s, ms, ns? That's especially true when we're working on a system using stuff that has timestamps defined in seconds (JWT expiry, possibly others?). A UTC string doesn't seem to have these issues, being self-describing, easy to interpret, etc. That said, I'm not wedded to this format because it's not clear to me how this data will be used by the client; likely there will be some parsing and formatting required before displaying to the user.

platform/access/src/jsonrpc/middleware/injectAccessNode.ts
crimeminister: There's an existing `InjectEdges` middleware I used as a model. Happy to rename, but we should maybe pick a convention and update everything in a new PR?

platform/access/src/jsonrpc/middleware/injectAccessNode.ts
crimeminister: Fixed.

platform/access/src/jsonrpc/middleware/injectAccessNode.ts
crimeminister: Oh, I think you're suggesting that we use this naming patter just for middleware that injects a DO proxy stub into the context, and not more generally for other things?

platform/access/src/nodes/access.ts
crimeminister: Fixed!

packages/platform-middleware/jwt.ts
crimeminister: Is there a reason why `null` would be assigned to this value if has the initial value of `undefined`? Or is this more in the spirit of defensive programming?

platform/access/src/context.ts
crimeminister: Not every route uses the middleware to inject the client, only the routes that require talking to that service do. Since the context is shared by all routes, the values that are only initialized for some routes I assume have to optional. Perhaps there's a more idiomatic way to deal with this in `tRPC`, or should we just use the `InjectEdges` middleware for all routes?

crimeminister:
> > It is not JSON-RPC anymore. I think the request body snippets in `curl` directory are useless now.
> 
> I'll remove them. Do we have any other examples in docs or files to demonstrate usage that we can replace them with?

Removed.

platform/access/src/jsonrpc/middleware/injectAccessNode.ts
crimeminister: Renamed.

platform/access/package.json
```diff
@@ -34,10 +34,12 @@
     "@ethersproject/bytes": "5.7.0",
     "@ethersproject/random": "5.7.0",
     "@kubelt/platform-clients": "workspace:*",
+    "@kubelt/platform-middleware": "workspace:*",
     "@kubelt/urns": "workspace:*",
     "@trpc/server": "10.8.1",
     "do-proxy": "1.3.3",
     "jose": "4.11.0",
-    "typed-json-rpc": "1.1.0"
+    "typed-json-rpc": "1.1.0",
```
maurerbot, line 42: I don't think we need this package anymore?

maurerbot (changes requested):

platform/access/package.json
szkl: Nope.

szkl (approved):

maurerbot (approved):

Issue #1383, opened 2023-01-10
fix(passport): when authenticating for the first time using blockchain account we are missing the step of adding an edge


maurerbot:
Also, Add a q-comp field that discriminates custodial from non custodial (vault) accounts.

maurerbot:
closed by: #1400

PR #1386, opened 2023-01-11
feat(address): initial `vault` account

szkl:


apps/passport/app/routes/authenticate/google/callback.ts
```diff
@@ -30,7 +30,7 @@ export const loader: LoaderFunction = async ({ request }: LoaderArgs) => {
   const encoder = new TextEncoder()
   const hash = keccak256(encoder.encode(profile._json.email))
```
maurerbot, line 31: I think this is what is making the bundle size so large?

apps/passport/app/routes/authorize/index.tsx
```diff
@@ -38,27 +43,53 @@ export const loader: LoaderFunction = async ({ request, context }) => {
...
           },
-          cover: voucher.metadata?.cover,
         },
-      },
-      {
+        {
+          [PlatformJWTAssertionHeader]: jwt,
+        }
+      )
+    } else if (rparams.get('node_type') == NodeType.OAuth) {
```
maurerbot, line 66: let's discuss how to make an address profile helper

platform/address/src/context.ts
```diff
@@ -23,6 +23,7 @@ interface CreateInnerContextOptions
   Edges: Fetcher
   Access: Fetcher
   address?: DurableObjectStubProxy<CryptoAddress | OAuthAddress>
+  address3RN?: AddressURN
```
maurerbot, line 26: what is the diff?

platform/address/src/jsonrpc/middlewares/parse3RN.ts
```diff
@@ -12,7 +12,7 @@ export const parse3RN: BaseMiddlewareFunction<Context> = async ({
   ctx,
   next,
 }) => {
-  const header = ctx.req?.headers.get('X-3RN')
+  const header = ctx.address3RN || ctx.req?.headers.get('X-3RN')
```
maurerbot, line 15: let's create a header const somewhere like what we did with the assertion token?

apps/passport/app/routes/authenticate/google/callback.ts
szkl: I haven't measured it yet. We'll find out soon.

platform/address/src/context.ts
szkl: This is supposed to take precedence over the request header so server-side calls can override the header value easily. It'd be better to put the header value into the context object from the beginning of the request handling, though.

platform/address/src/jsonrpc/methods/initVault.ts
```diff
@@ -0,0 +1,45 @@
...
+  await vaultNode.storage.put('privateKey', vault.privateKey)
+
+  const address3RN: AddressURN = `urn:threeid:address/${vault.address}?+node_type=crypto&addr_type=ethereum`
+  const caller = appRouter.createCaller({
+    ...ctx,
+    address3RN,
+  })
+
+  await Promise.all([
+    caller.setAccount(account),
```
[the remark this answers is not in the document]
szkl: Sure.

szkl:
> We need to also create a vault account for first time eth authentications to. Let's also add a qComp field `default=true`" on the address edge for:
> 
> 1. Vault account if social
> 2. Eth account if Eth

Why create a vault account if the signup address is a `CryptoAddress`?

Do you mean setting the `default=true` on the `CryptoAddress` node if it is a signup?

platform/address/src/context.ts
szkl: I'm checking if there would be any issues if the `ctx.addressURN` property is set as the header value. That name could be reusable.

szkl: The only reference that would be affected by keeping `ctx.addressURN` as its initial value would be instantiating the node clients which are wrapped in init functions. Those functions could get the base address URN to get a stub object. Though, I want to ask why not the full URN is used instead.

platform/address/src/jsonrpc/methods/initVault.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (!account) {
+    throw new Error('missing account')
+  }
+
+  const vault = Wallet.createRandom()
+  const vaultNode = await initCryptoNodeByName(
+    AddressURNSpace.urn(vault.address),
+    ctx.CryptoAddress
+  )
+  await vaultNode.storage.put('privateKey', vault.privateKey)
```
maurerbot, line 34: do we want to store a ref of the public key too?

szkl: I'm not sure.

maurerbot (approved):

szkl:
> Do we have address edge setup at every unit address now?
> 
> #1383
> 
> Merging and we that can be done in a new PR

I think so. I’ll check again.

szkl:
> > Do we have address edge setup at every unit address now?
> > #1383
> > Merging and we that can be done in a new PR
> 
> I think so. I’ll check again.

It looks okay right now. The edge will be created as long as `setAccount` is called when an address is created.

PR #1387, opened 2023-01-11
fix(clients): fix fetcher contexT

maurerbot:
wtf


szkl (dismissed):
`fetcher.fetch.bind(fetcher)` is an alternative.

szkl (approved):

PR #1401, opened 2023-01-13
feat(passport): OAuth Twitter

szkl:


platform/address/src/jsonrpc/validators/oauth.ts
```diff
@@ -72,18 +72,27 @@ export const GithubProfileSchema = z.object({
...
+  provider: z.literal(OAuthAddressType.Twitter),
+  id: z.number(),
+  name: z.string(),
+  screen_name: z.string(),
+})
+
 export const OAuthDataSchema = z.object({
   accessToken: z.string(),
-  refreshToken: z.string(),
+  accessTokenSecret: z.string().optional(),
```
szkl, line 84: This difference suggests the payload data schema should be a union.

apps/passport/app/routes/authenticate/twitter/callback.ts
```diff
@@ -0,0 +1,78 @@
...
+  await addressClient.setOAuthData.mutate({
+    accessToken,
+    accessTokenSecret,
+    profile: { ...profile, provider: OAuthAddressType.Twitter }
+  })
+
+  return authenticateAddress(address, account)
+}
+
+const authenticateAddress = async (
```
szkl, line 46: This function should be reusable in other routes.

szkl:
Setting up the account profile in the `/authorize` route needs to be able to work with different OAuth provider profile types.

PR #1446, opened 2023-01-23
fix(passport): Fix regression for MS avatar caching

betimshahini:
# Description

Previous authz PRs removed key functionality in MS integration callback handler. This reverts those changes.

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/apps/passport/app/utils/cacheImageToCF.server.ts
+++ b/apps/passport/app/utils/cacheImageToCF.server.ts
@@ -1,0 +1,36 @@
+export default async (
+  image_retrieval_url: string,
+  headers?: Record<string, string>
+): Promise<string> => {
+  const retrieveImageReq = fetch(image_retrieval_url, {
+    headers: headers,
+    method: 'get',
+  })
+  const retrievedImage = await retrieveImageReq
+
+  if (!retrievedImage.ok) {
+    if (retrievedImage.status === 404) {
+      //Valid error
+      console.error('No image found in at that URL.')
+      return ''
+    } else {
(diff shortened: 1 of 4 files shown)


apps/passport/app/utils/cacheImageToCF.server.ts
```diff
@@ -0,0 +1,36 @@
+export default async (
```
maurerbot, line 1: redundant

maurerbot (approved):

apps/passport/app/utils/cacheImageToCF.server.ts
maurerbot: j/k

platform/address/src/jsonrpc/validators/profile.ts
```diff
@@ -92,6 +92,7 @@ export const MicrosoftRawProfileSchema = z.object({
   family_name: z.string(),
   email: z.string(),
   picture: z.string(),
+  threeidImageUrl: z.string().optional(),
```
maurerbot, line 95: why a separate field?

maurerbot (changes requested):

platform/address/src/jsonrpc/validators/profile.ts
maurerbot: we should jsut use the profile field and when we query for the address profile just repeat the process

maurerbot (approved):
closing but new ticket to move address profile to address node

PR #1529, opened 2023-01-31
feat(address): oauth refresh token

szkl:


platform/address/.dev.vars.example
```diff
@@ -7,4 +7,20 @@ APIKEY_ALCHEMY_POLYGON =
 ALCHEMY_POLYGON_NETWORK = 
 
 APIKEY_MORALIS =
-MORALIS_STREAM_ID =
\ No newline at end of file
+MORALIS_STREAM_ID =
```
maurerbot, line 10: delete moralis

szkl: https://github.com/kubelt/kubelt/pull/1559

platform/address/src/jsonrpc/methods/getAddressProfile.ts
```diff
@@ -93,23 +93,23 @@ export const getAddressProfileMethod = async ({
       }
     }
     case OAuthAddressType.Google: {
-      const oAuthNode = new GoogleAddress(nodeClient)
+      const oAuthNode = new GoogleAddress(nodeClient, ctx)
```
maurerbot, line 96: should we pass context to all the typed nodes to be consistent?

szkl: It was initially an `OAuthAddress` constructor parameter but then it was just used to just to set a couple of related object properties like `clientId` and `clientSecret`. I didn't see the need for having the context in full.

maurerbot: ok

maurerbot (approved):

PR #1556, opened 2023-02-02
feat(address): fetch oauth profile

### LINKED ISSUE #1455: chore(address): getAddressProfile() calls for OAuth addresses should retrieve from source
To ensure freshness of external (OAuth provider) data, the `getAddressProfile()` calls for each provider's Address nodes should be tweaked to do call-time fetches to the OAuth provider. 

- [x] This fetch should be cached outbound using the fetcher `cacheKey` property to ensure some caching (5 min?) for frequently requested data.
- [x] For the MS provider, there's an additional consideration that the avatar image would need to be fetched as well (with their issued access token) can cached using the CF image service, and for the CF Image cache URL to be returned instead. 
- [x] In the absence of any profile images, the gradient should be returned.
- [x] The shape of the `getAddressProfile()` should be normalized to return the same shape. This way Passport's authorize route can be simplified.

szkl:
closes #1455


platform/address/src/nodes/oauth.ts
```diff
@@ -29,6 +29,28 @@ export default class OAuthAddress {
     return this.node.storage.put('data', data)
   }
 
+  async getProfile(): Promise<OAuthData['profile'] | undefined> {
+    try {
+      const response = await fetch(
+        this.getUserInfoURL(),
+        {
+          cf: {
+            cacheTtl: 5 * 60,
```
maurerbot, line 38: any rational behind cache time?

szkl: Not really. Just the initial note in #1455

platform/address/src/nodes/apple.ts
```diff
@@ -19,12 +19,12 @@ export default class AppleAddress extends OAuthAddress {
     this.clientSecret = ctx.SECRET_APPLE_OAUTH_CLIENT_SECRET
   }
 
-  async getProfile(): Promise<OAuthAppleProfile> {
+  async getProfile<T = OAuthAppleProfile>(): Promise<T> {
```
maurerbot, line 22: apple the exception since there is no profile api call?

platform/address/src/nodes/twitter.ts
```diff
@@ -4,11 +4,11 @@ import Address from './address'
 import OAuthAddress from './oauth'
 
 export default class TwitterAddress extends OAuthAddress {
-  async getProfile(): Promise<OAuthTwitterProfile> {
+  async getProfile<T = OAuthTwitterProfile>(): Promise<T> {
     const data = await this.getData()
```
maurerbot, line 8: is twitter another exception because of paywall coming?

platform/address/src/nodes/apple.ts
szkl: Yeah. Hence, the method override.

platform/address/src/nodes/twitter.ts
szkl: I kept this method override because the super method is based on OAuth 2.0 providers. The current Twitter OAuth client is v1 and requires more development to make a request. So, another reason to keep the method override is parking `twitter` as an OAuth provider.

maurerbot (dismissed):

szkl:
> Since this was split off from the oauth refresh token ticket, why isn't it being done under #1455 along with the other items there?

I did this to test #1529. It covers the base functionality requirement of #1455. It is possible to address an issue with multiple requests.

PR #1588, opened 2023-02-06
chore(galaxy, profile): nft optimization

### LINKED ISSUE #1494: chore(galaxy): optimize blockchain APIs

### LINKED ISSUE #1607: fix(profile): gallery settings should use multi crypto addresses

poolsar42:
# Description

Optimizing nft calls

It doesn't use multiple addresses in the client. Instead it sends accountURN to the galaxy and fetches all addresses from this accountURN in the galaxy.

- Closes #1494
- Closes #1607

- [x] attach chain and network on nfts and contracts in galaxy
- [x] do it in alchemy client(?) package
- [x] separate calls to different chains parts in promise.alls (instead of calling all contracts on all networks)
- [x] make a brief note on where sort.xyz could be useful for us
- [ ] use https://www.npmjs.com/package/async where needed
- [x] make calls algorithmically more efficient in terms

## Type of change

- Optimization

# How Has This Been Tested?

# Checklist:


apps/profile/app/components/accounts/PfpNftModal.tsx
```diff
@@ -35,7 +35,7 @@ const PfpNftModal = ({
   setCollection,
 }: PfpNftModalProps) => {
   const displayText = `Looks like ${
-    displayName ?? account
+    displayName ?? addresses[0]
```
maurerbot, line 38: when are we going to support multiple addresses? is that pr coming next?

maurerbot, line 38: actually, I don't think we will ever not have a display name?

apps/profile/app/routes/$type.$address/collections/$collection.tsx
```diff
@@ -92,11 +92,10 @@ const CollectionForProfileRoute = () => {
         })
       }}
       loadingConditions={loading || refresh}
-      account={tempTargetAddress} // #TODO: replace with list of visible addresses
       isModalNft={true}
       isOwner={isOwner}
       displayText={`Looks like ${
-        displayName ?? tempTargetAddress
+        displayName ?? targetAddresses[0]
```
maurerbot, line 98: same here. will we ever not have display name?

apps/profile/app/routes/$type.$address/collections/index.tsx
```diff
@@ -89,12 +79,11 @@ const ProfileRoute = () => {
       isModal={false}
       loadingConditions={loading || refresh}
       nfts={loadedNfts}
-      account={tempTargetAddress}
       pfp={(pfp as any).image as string}
       isOwner={isOwner}
       filters={true}
       displayText={`Looks like ${
-        displayName ?? tempTargetAddress
+        displayName ?? targetAddresses[0]
```
maurerbot, line 86: again here.

apps/profile/app/routes/account/gallery.tsx
```diff
@@ -478,7 +475,7 @@ const Gallery = () => {
             name="gallery"
             value={JSON.stringify(curatedNfts)}
           />
-          <input type="hidden" name="address" value={tempTargetAddress} />
+          <input type="hidden" name="address" value={targetAddresses[0]} />
```
maurerbot, line 478: is this needed? or should it be targetAddresses?

platform/galaxy/src/schema/resolvers/account.ts
```diff
@@ -108,19 +109,38 @@ const accountResolvers: Resolvers = {
...
+      // Validation
+      if (gallery) {
+        const validator = await validOwnership(
+          gallery as Gallery,
+          env,
+          connectedAddresses
+        )
+
+        return gallery.filter((nft) => {
+          return validator.get(nft.contract)?.includes(nft.tokenId)
```
maurerbot, line 131: here are we checking when we return the gallery that the tokens still belong to this account?

maurerbot: we should update the gallery (fire and forget, don't await) with filtered out list

maurerbot (changes requested):

platform/galaxy/src/schema/resolvers/account.ts
poolsar42: > here are we checking when we return the gallery that the tokens still belong to this account?

yes

apps/profile/app/routes/$type.$address/collections/$collection.tsx
poolsar42: > same here. will we ever not have display name?

I think there are some cases (e.g. Github) when there's could be no displayName set. 
The good question is: is there a targetAddress then? I can't answer that right away, but probably it should be set.
I will investigate this and ping you after

maurerbot: I think we always set the display name to at least the sub?

cc @betimshahini

poolsar42: > 

So far this is done for microsoft, but I'm not sure about others

poolsar42: @maurerbot I think [this](https://github.com/proofzero/rollupid/blob/main/apps/passport/app/utils/authenticate.server.ts#L88) is where displayname is being set. In this case I removed lines with conditional `targetAddresses[0]`

maurerbot (changes requested):
remove gallery items if filtered out

platform/galaxy/src/schema/resolvers/account.ts
```diff
@@ -273,15 +278,20 @@ const accountResolvers: Resolvers = {
...
+          gallery,
           env,
           connectedAddresses
         )
 
-        return gallery.filter((nft) => {
-          return validator.get(nft.contract)?.includes(nft.tokenId)
-        })
+        if (gallery.length !== filteredGallery.length) {
+          await accountClient.setGallery.mutate({
```
maurerbot, line 288: probably don't need to await

maurerbot (changes requested):

maurerbot (approved):
thanks

PR #1590, opened 2023-02-07
feat(profile): Update dashboard to new design

Cosmin-Parvulescu:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

- Closes #1505 

## Type of change


Cosmin-Parvulescu:
> Does not seem to work. Got the following errors, and a pepe page. Tried Github and wallet accounts.
> 
> image
> 
> image

Interesting. This might be because you have edge data from before `createdTimestamp` was added to the `edge` table. I could add an additional filter (we also have one for clientId as there are cases where it was not set), but maybe the wisest choice would be to assume that `createdTimestamp` has to be there.

platform/account/src/jsonrpc/methods/getApps.ts
```diff
@@ -0,0 +1,92 @@
...
+      // We only want edges that start at the provided account node.
+      id: input.account,
+      // We only want edges that link to Access nodes (sessions).
+      tag: EDGE_ACCESS,
+      // Account -> Access edges indicate session ownership.
+      dir: Graph.EdgeDirection.Outgoing,
+    },
+  })
+
+  const mappedEdges = edgesResult.edges
```
Cosmin-Parvulescu, line 57:
```suggestion
  const mappedEdges = edgesResult.edges
      // Exclude edges from db before createdTimestamp 
      // existed so as to not error out
      .filter((edge) => edge.createdTimestamp)
```

Cosmin-Parvulescu:
> > Managed to reproduce the issue locally by deleting the `createdTimestamp` entries. Added a suggestion for a fix: https://github.com/proofzero/rollupid/pull/1590/files#r1102500581 though I'm not sure we should go through with it because we wouldn't expect a state in which that column is not set for that table.
> 
> Knowing what the error condition is at least gets us moving (with everyone needing to delete/recreate all their local state. Acceptable compromise for now) However, the whole page erroring out isn't great. Can we add error/catch boundary to the panel?

As far as I am aware, the panel is just a component inside the page (i.e. not part of the routing structure) so erroring out would still error the entire `/dashboard` page (not only the panel), so only the page layout would remain. I think that's an interesting suggestion. Should I add that as part of this PR?

Cosmin-Parvulescu:
> > As far as I am aware, the panel is just a component inside the page (i.e. not part of the routing structure) so erroring out would still error the entire /dashboard page (not only the panel), so only the page layout would remain. I think that's an interesting suggestion. Should I add that as part of this PR?
> 
> Use a remix fetcher so you can handle things in the panel or move panel into route so it's failure is isolated.

Remix fetcher is indeed a strategy; I am not sure moving the panel into its own route is a valid one. 

I will move fwd: with your suggestion of using a fetcher. 
That also means we might need:
1 - Loading state for the panel
2 - Error state for the panel
3 - Empty state for the panel

Let me know your thoughts on this.

apps/profile/app/routes/account/dashboard.tsx
```diff
@@ -72,103 +53,223 @@ export default function Welcome() {
...
-                  <div className="roadmap-next__check mt-1 flex justify-center items-top">
-                    <img src={stepSoon} alt="Rollup Step" />
-                  </div>
+          <div className="border shadow flex-1 flex flex-col rounded-lg">
+            <div className="bg-[#F9FAFB] flex items-center py-5 px-8 rounded-t-lg">
+              <Text size="sm" weight="medium" className="text-gray-500 flex-1">
+                APPLICATION
+              </Text>
+              <Text size="sm" weight="medium" className="text-gray-500 flex-1">
+                APPROVED
```
maurerbot, line 204: let's change to AUTHORIZED?

maurerbot (approved):

PR #1611, opened 2023-02-08
fix(access): add rcomps for better queries

maurerbot:
closes: #1604


packages/urns/access.ts
```diff
@@ -1,6 +1,12 @@
 import { createRollupIdURNSpace, RollupIdURN } from './index'
 
+export type AccessRComp = {
+  client_id?: string
+  grant_type?: string
```
szkl, line 5: I think camel casing fits better for internal use.

maurerbot: fixed

PR #1614, opened 2023-02-08
fix(profile): compare the expiration time in milliseconds

szkl:
# Description

The token expiration time is in seconds and `Date.now()` returns time in
milliseconds. The comparison results in expiration always because of that.

## Type of change

# How Has This Been Tested?

- Manual


szkl:
I'll search for any other occurrences of the same comparison expression.

maurerbot (commented):
https://media.giphy.com/media/ToMjGpkNsMxlvrj8DYc/giphy.gif

PR #1618, opened 2023-02-09
Chore/console users

poolsar42:
# Description

User Authorizations for `$clientId/users` (which is out of scope) and dashboard in console.

- Closes #1506 

# How Has This Been Tested?

# Checklist:


platform/starbase/src/jsonrpc/methods/getAuthorizedAccounts.ts
```diff
@@ -0,0 +1,68 @@
...
+    },
+    // set limit to not query the whole db
+    opt: { limit: 8 },
+  })
+
+  const mappedEdges = edgesResult.edges
+    .filter(
+      (edge) =>
+        (edge.dst.rc as AccessRComp).client_id &&
+        (edge.dst.rc as AccessRComp).client_id === input.client
```
szkl, line 54: The property name is going to change [here](https://github.com/proofzero/rollupid/pull/1603/files#diff-47d3e8099987fbce3da7904ec39ae4e868bb9d145d10f2ca0c4e3fd6c1d3008fR4).

szkl, line 54: It'd be great if `getEdges` can take type parameters which would eliminate (repetitive) type assertions. Another alternative would be a conditional expression (or an if statement) to fail early or ensure type safety on the property.

platform/starbase/src/jsonrpc/methods/getAuthorizedAccounts.ts
```diff
@@ -0,0 +1,68 @@
...
+    )
+    .map((edge) => {
+      const timestamp = new Date(
+        (edge.createdTimestamp as string) + ' UTC'
```
szkl, line 58: The time zone information should be in the data.

Not in the scope of the issue of this pull request.

platform/starbase/src/jsonrpc/methods/getAuthorizedAccounts.ts
```diff
@@ -0,0 +1,68 @@
...
+      ).getTime()
+      const accountURN = edge.src.urn
+
+      return { accountURN, timestamp }
+    })
+    // Order in ascending order
+    .sort((a, b) => a.timestamp - b.timestamp)
```
szkl, line 65: `Date` objects are comparable and can be applied to arithmetic operators.

szkl, line 65: Is it possible to get the data sorted from the source?

platform/starbase/src/jsonrpc/methods/getAuthorizedAccounts.ts
poolsar42: will change it on rebase

platform/starbase/src/jsonrpc/methods/getAuthorizedAccounts.ts
poolsar42: For this pr I'll make conditional type check to avoid repetitions

platform/starbase/src/jsonrpc/methods/getAuthorizedAccounts.ts
maurerbot: edges is broken rn. We will be able to do eliminate these javascript filters when we get local d1 support

platform/starbase/src/jsonrpc/methods/getAuthorizedAccounts.ts
maurerbot: rn it's fixed by created timestamp in asc order. same as above, we will be able to fix when d1 support locally and introduce orm for queries

platform/starbase/src/jsonrpc/methods/getAuthorizedAccounts.ts
szkl: This change won't be needed.

poolsar42: tbh I think now it's easier to keep it this way

Issue #1612, opened 2023-02-08
fix(apps): signout flow


maurerbot:
- [ ] when signing out of profile we should sent you to a new signout route in passport to also sign out of passport
- [ ] investigate if when cookies exists but data is empty (b/c we signed out) we don't assume data is there just because the cookie is there

PR #1633, opened 2023-02-10
fix(platform): signout

maurerbot:
closes: #1612

### DIFF
--- a/apps/console/app/utilities/session.server.tsx
+++ b/apps/console/app/utilities/session.server.tsx
@@ -5,11 +5,7 @@
 import invariant from 'tiny-invariant'
 import * as jose from 'jose'
 import type { JWTPayload } from 'jose'
-import {
-  createCookieSessionStorage,
-  // createCloudflareKVSessionStorage,
-  redirect,
-} from '@remix-run/cloudflare'
+import { createCookieSessionStorage, redirect } from '@remix-run/cloudflare'
 
 // @ts-ignore
 invariant(DEPLOY_ENV, 'DEPLOY_ENV must be set')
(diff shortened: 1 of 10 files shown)


apps/passport/app/session.server.ts
```diff
@@ -10,15 +10,15 @@ import type { JWTPayload } from 'jose'
 
 // USER PARAMS
 
-const getUserSessionStorage = (env: Env) => {
+const getUserSessionStorage = (env: Env, MAX_AGE = 60 * 60 * 24 * 90) => {
```
szkl, line 13: Prefer to have it defined in a `constants` module.

apps/passport/server.js
```diff
@@ -12,7 +12,10 @@ export function parseParams(request) {
     clientId,
     state,
     redirectUri,
-    scope,
+    scope:
+      scope && scope.trim() !== 'null' && scope.trim() !== ''
+        ? scope.split(',')
```
szkl, line 17: It is okay to just call `split` without trimming.

apps/passport/server.js
```diff
@@ -12,7 +12,10 @@ export function parseParams(request) {
     clientId,
     state,
     redirectUri,
-    scope,
+    scope:
+      scope && scope.trim() !== 'null' && scope.trim() !== ''
```
szkl, line 16: What's the possibility of getting a string value `'null'`? Should be eliminated if it is possible.

maurerbot: still seeing it in this branch

apps/console/app/utilities/session.server.tsx
```diff
@@ -95,21 +62,19 @@ export async function logout(request: Request) {
...
-    throw redirect(PASSPORT_URL)
+    throw redirect(`${PASSPORT_URL}/signout`)
   }
   if (jwt) {
     const parsedJWT = parseJwt(jwt)
-    if (parsedJWT.exp < Date.now() / 1000) {
-      throw await destroyUserSession(session)
+    const exp = parsedJWT?.exp
+    if (exp && exp < Date.now() / 1000) {
+      throw redirect(`${PASSPORT_URL}/signout`)
```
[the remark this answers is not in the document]
maurerbot: yes, but the cookie could be alive and I'd like to delete it.

apps/passport/server.js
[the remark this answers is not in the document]
maurerbot: iirc scopes are always comma deliminated in OAuth?

apps/passport/app/session.server.ts
maurerbot: removed calc b/c I prefer not to maintain seperately rn and save the computation

PR #1641, opened 2023-02-13
fix(images): microsoft pfp cache upload

### LINKED ISSUE #1639: fix(platform): Fix Blob image uploads in tRPC interface
Passport's `cacheImageToCF` is used during the Microsoft authentication to retrieve an avatar from MS Graph and cache it for later retrieval. The caching is done by the Images service, where the blob of the avatar is sent through tRPC. Currently, tRPC does not support this operation, and on the receiving side of that call, the value is discarded. For this reason, Microsoft logins are currently throwing errors and user's can't log in.

To fix, for the short term, the image Blob will need to be encoded into the JSON object (that tRPC supports), through a conversion to a data URL (base64 encoding with MIME type set) on the sending side (Passport), and conversion back to Blob on the receiving side (Images service).

Reference for methods to do b64->Blob and vice-versa: https://ionic.io/blog/converting-a-base64-string-to-a-blob-in-javascript

poolsar42:
# Description

- Closes #1639 

# Checklist:


apps/passport/app/utils/cacheImageToCF.server.ts
```diff
@@ -25,14 +27,33 @@ export default async (
...
+  } | null = null
+
+  try {
+    cfUploadRes = await fetch(uploadURL, {
+      method: 'POST',
+      body: formData,
+    }).then((res) => res.json())
+
+    if (!cfUploadRes?.success || !cfUploadRes.result?.variants)
+      throw new Error('Upload unsuccessful')
```
szkl, line 49:
```suggestion
      throw new Error('Upload failed')
```

apps/passport/app/utils/cacheImageToCF.server.ts
```diff
@@ -25,14 +27,33 @@ export default async (
...
+  try {
+    cfUploadRes = await fetch(uploadURL, {
+      method: 'POST',
+      body: formData,
+    }).then((res) => res.json())
+
+    if (!cfUploadRes?.success || !cfUploadRes.result?.variants)
+      throw new Error('Upload failed')
+  } catch (ex) {
+    console.error('Could not send upload image blob to Image cache.', ex)
```
szkl, line 51: Should it return early if an error occurred?

poolsar42: yes I think so

apps/passport/app/utils/cacheImageToCF.server.ts
```diff
@@ -11,6 +11,8 @@ export default async (
   })
   const retrievedImage = await retrieveImageReq
 
+  const imgFile = await retrievedImage.blob()
```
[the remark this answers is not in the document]
poolsar42: Done

szkl (commented):
I think it'd be much easier to read and avoid existence check if the main logic in the change set is in the `try` clause.

poolsar42:
> I think it'd be much easier to read and avoid existence check if the main logic in the change set is in the `try` clause.

I think that's a good point. Did something like this.

apps/passport/app/utils/cacheImageToCF.server.ts
```diff
@@ -11,28 +11,32 @@ export default async (
   })
   const retrievedImage = await retrieveImageReq
 
-  if (!retrievedImage.ok) {
-    if (retrievedImage.status === 404) {
```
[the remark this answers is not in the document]
poolsar42: Okay. I think if we need to have a specific 404 check here, we still can check for other problems at the same place, so I will keep this if statement.

Issue #1692, opened 2023-02-16
fix(profile): when I have an existing session on profile and login with a different account to console the nav to profile from console is broken


maurerbot:
when I have an existing session on profile and login with a different account to console the nav to profile from console shows my existing profile session user.

one idea to fix this is: navigating to profile from console should take me to profile /signout and redirect my back to profile with the right account

szkl:
Why shouldn't the passport application redirect to the console if the user is already logged in?

maurerbot:
> Why shouldn't the passport application redirect to the console if the user is already logged in?

That isn't the issue. It's the expectation we are setting in the UI that it's "one app" and that when I click profile from console I should be in the same session context but in some cases I am not. I am user A in console and user B in profile.

szkl:
> one idea to fix this is: navigating to profile from console should take me to profile /signout and redirect my back to profile with the right account

Do you mean hitting profile/logout regardless of the current user?

szkl:
> 

I think it could be a more controlled process if the console application redirected to the passport requesting a login to the profile application with the specified account. I'm assuming the passport application will be able to tell the currently logged-in profile user.

PR #1708, opened 2023-02-17
fix(passport): console params session should not expire

### LINKED ISSUE #1695: fix(passport): console params session should not expire
if you leave the authentication screen for more than 5 minutes and you login you will be taken to console but the branding is based on client id. Because of this we should not expire console params but make sure we override them with new client id

Cosmin-Parvulescu:
# Description

- Added 400 days expiration date for cookie
- If context has new consoleParams that have a different clientId, replaces console props cookie

- Closes #1695 
- Depends on #1692 

## Type of change

# How Has This Been Tested?

1 - Go to Profile App A
2 - Get redirected to Passport
3 - Get Profile App A branding
4 - Wait 6 minutes
5 - Go to Passport app
6 - See Profile App A branding
7 - Go to Profile App A
8 - Log out
9 - Go to Profile App B
10 - Get redirected to Passport
11 - See Profile B branding

### DIFF
--- a/apps/passport/app/session.server.ts
+++ b/apps/passport/app/session.server.ts
@@ -84,7 +84,10 @@ export async function logout(request: Request, redirectTo: string, env: Env) {
 
 const getConsoleParamsSessionStorage = (
   env: Env,
-  MAX_AGE = 300 // 5 minutes
+  // https://developer.chrome.com/blog/cookie-max-age-expires/
+  // As of Chrome release M104 (August 2022) cookies can no longer
+  // set an expiration date more than 400 days in the future.
+  MAX_AGE = 34_560_000
 ) => {
   return createCookieSessionStorage({
     cookie: {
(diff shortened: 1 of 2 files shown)


apps/passport/app/session.server.ts
```diff
@@ -84,7 +84,10 @@ export async function logout(request: Request, redirectTo: string, env: Env) {
 
 const getConsoleParamsSessionStorage = (
   env: Env,
-  MAX_AGE = 300 // 5 minutes
+  // https://developer.chrome.com/blog/cookie-max-age-expires/
+  // As of Chrome release M104 (August 2022) cookies can no longer
+  // set an expiration date more than 400 days in the future.
+  MAX_AGE = 34_560_000
```
maurerbot, line 90: underscores work?

also, have we tested that with coming back to passport with different client ids that the cookie get's new data?

Cosmin-Parvulescu: - They should work with our bundlers: https://v8.dev/features/numeric-separators ;
- I tested it with two apps (algo in "How Has This Been Tested?"); coming from app B via its redirect replaces the cookie's data with the new app's props.

maurerbot (dismissed):

maurerbot (approved):

PR #1709, opened 2023-02-17
feat(profile): Applications listing and revocation modal

### LINKED ISSUE #1669: feat(profile): Applications listing and revocation modal
Following the Figma design in Applications > Applications view, implement:
- [ ] From Applications mockup: List application view
- [ ] From "Edit Access" mockup: modal (read-only - links/buttons aren't clickable other than Cancel)

Cosmin-Parvulescu:
# Description

- [x] Add app list route
- [x] Enable app list menu item
- [x] Link View All from dashboard apps to route
- [x] Application list items (needs scope data) 🚫
- [x] Modal
- [x] Fix hydration difference between timestamp on client / server...

- Closes #1669 

## Type of change

# How Has This Been Tested?


apps/profile/app/routes/account/dashboard.tsx
```diff
@@ -29,7 +28,7 @@ export default function Welcome() {
   const appFetcher = useFetcher()
 
   useEffect(() => {
-    appFetcher.load('/account/dashboard/apps')
+    appFetcher.load('/api/apps')
```
maurerbot, line 31: should we use the same defer method here like what we're doing in console?

platform/access/src/jsonrpc/methods/getAppScopes.ts
```diff
@@ -0,0 +1,78 @@
...
+  const uniqueScopes = scopes.filter((v, i, a) => a.indexOf(v) === i)
+
+  // Add implicit openid scope
+  uniqueScopes.push('scope://rollup.id/openid')
+
+  // Generate an array of [{
+  //   permission: 'read' | 'write' | 'root' | ... based on scope structure,
+  //   scope: 'scope'
+  // }]
+  const castScopes = uniqueScopes
```
maurerbot, line 46: is this a presentation concern or something we want to do everywhere?

Cosmin-Parvulescu: ATM it's used only in applications / index, I'm not sure if we'll use the same structure everywhere. You guys mentioned doing all transforms inside the worker and keeping Galaxy light, which is why they're here. @betimshahini ?

platform/access/src/jsonrpc/methods/getAppScopes.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  const { tokenIndex, tokenMap } = await accessNode.class.getTokenState()
+
+  // Get a map of all the scopes
+  // in all authorizations
+  const tokens = tokenIndex.map((t) => tokenMap[t])
+  const scopes = tokens.flatMap((t) => t.scope)
+
+  // Filter for unique scopes
+  const uniqueScopes = scopes.filter((v, i, a) => a.indexOf(v) === i)
```
szkl, line 37: Just a note. Each iteration in `filter()` also does another loop by calling `indexOf()`.

Cosmin-Parvulescu: Hmm, would be better if we used a set maybe.

apps/profile/app/routes/account/dashboard.tsx
maurerbot: the promise retruned to the client would throw an exception so an internal catch boundry can be used

Cosmin-Parvulescu: I think it still effects the entire dashboard context, not just the application panel of the dashboard

maurerbot: depends on how the components are structured but you can also `.catch` the client promise too to manage error in that panel.

platform/access/src/jsonrpc/methods/getAppScopes.ts
```diff
@@ -0,0 +1,78 @@
...
+  })
+)
+
+export const getAppScopesMethod = async ({
+  input,
+  ctx,
+}: {
+  input: z.infer<typeof GetAppScopesInput>
+  ctx: Context
+}): Promise<z.infer<typeof GetAppScopesOutput>> => {
```
szkl: I find the pattern [here](https://github.com/proofzero/rollupid/blob/c698d90a6ab1fe8ab418a26762c9847dc25110a4/platform/access/src/jsonrpc/methods/verifyToken.ts#L9-L28) easier to read and edit.

platform/access/src/jsonrpc/methods/getAppScopes.ts
```diff
@@ -0,0 +1,78 @@
+import { z } from 'zod'
+import { Context } from '../../context'
+import { initAccessNodeByName } from '../../nodes'
+import { inputValidators } from '@kubelt/platform-middleware'
+import { AccountURNSpace } from '@kubelt/urns/account'
+import { scope, SCOPES } from '@kubelt/security/scopes'
+
+export const GetAppScopesInput = z.object({
+  accountURN: inputValidators.AccountURNInput,
+  clientId: z.string(),
```
szkl: Would validate an empty string.

szkl (changes requested):

apps/profile/app/routes/account/dashboard.tsx
Cosmin-Parvulescu: @betimshahini do you think it's a good idea to create a new issue instead of handling it in this PR?

platform/access/src/jsonrpc/methods/getAppScopes.ts
Cosmin-Parvulescu: Added a `.min(1)` validation, as suggested by the `.nonempty()` deprecation docs.

platform/access/src/jsonrpc/methods/getAuthorizedAppScopes.ts
```diff
@@ -5,74 +5,89 @@ import { inputValidators } from '@kubelt/platform-middleware'
 import { AccountURNSpace } from '@kubelt/urns/account'
 import { scope, SCOPES } from '@kubelt/security/scopes'
 
-export const GetAuthorizedAppScopesInput = z.object({
+export const GetAuthorizedAppScopesMethodInput = z.object({
   accountURN: inputValidators.AccountURNInput,
-  clientId: z.string(),
+  clientId: z.string().min(1),
```
szkl, line 10: That's cheating.

szkl: Joke aside. Might wanna consider an actual validator definition for the client id. We haven't defined one yet as far as I can tell.

Cosmin-Parvulescu: > That's cheating.

It's what they recommend instead of deprecated nonempty 😁

> Joke aside. Might wanna consider an actual validator definition for the client id. We haven't defined one yet as far as I can tell.

We use clientId as `z.string()` throughout so it would be an overreaching new Issue imho :) I agree it'd be best to have a specific validator as with the URN one.

szkl: The current implementation would return an empty list but that would be semantically not correct if the given client id is not valid. These could be filed into the backlog if considered as needed.

Cosmin-Parvulescu: Refactoring this would touch an additional 12 files not related to the PR so I still think it should be part of a separate Issue. @betimshahini ? :)

apps/profile/app/routes/account/dashboard.tsx
maurerbot: new pr. this is fine for now.

maurerbot (approved):

szkl (approved):

Issue #1715, opened 2023-02-17
chore(gha): Make chromatic jobs conditional on component or design system changes


betimshahini:
To reduce wait time for PR checks as well as avoid all together Chromatic jobs running for non-component changes, make GHA workflow conditional on running only on changes to files in:
- [x] apps/console/app/components
- [x] apps/profile/app/components
- [x] apps/passport/app/components
- [x] packages/design-system

PR #1722, opened 2023-02-17
fix(repo): storybook pr #1715

maurerbot:


szkl (dismissed):

.gitignore
```diff
@@ -227,3 +227,7 @@ act.secrets
 # api-extractor
 temp
 storybook-static
+
+# secrets
+**/*.p8
+
```
szkl, line 233: I don't think it is a good idea to ignore it. It could go unnoticed.

maurerbot: > I can send that script accepting CLI options for the key file path and the client id.

If you send a script with cli options then we can drop this

szkl (approved):

.gitignore
szkl: on it

szkl: https://github.com/proofzero/rollupid/pull/1725

PR #1769, opened 2023-02-22
fix(access): no need for authentication edges

maurerbot:


platform/access/src/jsonrpc/methods/exchangeToken.ts
```diff
@@ -99,14 +100,10 @@ const handleAuthenticationCode: ExchangeTokenMethod<
   )) as AccountURN
```
szkl, line 100: This kind of type assertion can lead to errors go unnoticed.

platform/access/src/jsonrpc/methods/exchangeToken.ts
```diff
@@ -99,14 +100,10 @@ const handleAuthenticationCode: ExchangeTokenMethod<
...
   const scope: Scope = (await authorizationNode.storage.get('scope')) || []
 
-  const name = `${AccountURNSpace.decode(account)}@${clientId}`
+  const name = `${AccountURNSpace.decode(account)}@${account}`
```
szkl, line 103: `account` could be nullish.

PR #1802, opened 2023-02-28
chore(platform): error handling development protocol

szkl:
# Description

TBD

## Type of change

# How Has This Been Tested?

- Manual introspection


szkl:
Added `chore(access): refresh token exchange error handling` section in the
description.

apps/passport/app/routes/token.tsx
```diff
@@ -21,33 +23,50 @@ export const action: ActionFunction = async ({ request, context }) => {
...
+        {
+          code: error.data.code,
+          message: error.shape.message,
+        },
+        {
+          status: error.data.httpStatus,
+        }
+      )
+    } else {
+      throw error
```
szkl, line 69: This is still going to abruptly terminate the request process.

apps/profile/app/routes/account/dashboard/apps.tsx
```diff
@@ -3,23 +3,28 @@ import { getAuthorizedApps } from '~/helpers/profile'
 import { requireJWT } from '~/utils/session.server'
 
 export const loader: LoaderFunction = async ({ request }) => {
-  const jwt = await requireJWT(request)
```
szkl, line 6: Unhandled errors will cause page load to result in a HTTP 500.

apps/profile/app/routes/account/dashboard/apps.tsx
```diff
@@ -3,23 +3,28 @@
...
 
-    return {
-      error: true,
+      return {
+        error: true,
+      }
     }
+  } catch (error) {
+    console.error({ error })
+    return null
```
szkl, line 28: Not sure if there is anything to be done regarding error handling here. Perhaps looking up the error code would be helpful.

apps/profile/app/utils/session.server.tsx
```diff
@@ -88,14 +88,9 @@ export async function requireJWT(request: Request, headers = new Headers()) {
       const token = await fetch(PASSPORT_TOKEN_URL, {
         method: 'post',
         body: form,
-      }).catch((err) => {
-        console.error('failed to refresh token', err)
-        throw redirect('/signout')
```
szkl, line 93: The response has to be non-200 for this catch handler to be invoked.

apps/passport/app/routes/token.tsx
szkl: Needs a generic error response body and status code.

platform/access/src/jsonrpc/methods/exchangeToken.ts
```diff
@@ -235,7 +237,18 @@ const handleRefreshToken: ExchangeTokenMethod<RefreshTokenInput> = async ({
   const name = `${AccountURNSpace.decode(account)}@${clientId}`
   const accessNode = await initAccessNodeByName(name, ctx.Access)
 
-  await accessNode.class.verify(refreshToken)
+  try {
+    await accessNode.class.verify(refreshToken)
+  } catch (cause) {
+    if (cause instanceof jose.errors.JOSEError) {
+      const code = 'UNAUTHORIZED'
```
szkl, line 244: tRPC error codes are not defined as an enumeration. The pre-defined [codes](https://github.com/trpc/trpc/blob/19c7c27d8ed548b2bc7ab5eb921fd179143aa230/packages/server/src/rpc/codes.ts) also set the response status code.

apps/passport/app/routes/token.tsx
```diff
@@ -21,33 +23,50 @@ export const action: ActionFunction = async ({ request, context }) => {
...
+    return json(result, {
+      //spec adherence and general good practice
+      headers: {
+        'Cache-Control': 'no-store',
+        Pragma: 'no-cache',
+      },
+    })
+  } catch (error) {
+    if (error instanceof TRPCClientError) {
+      console.error(error.toString())
```
szkl, line 58: Calling the `toString()` method on an actual error instance logs the error message's name and title. It should be used for error messages that can happen frequently and shouldn't cause too much disruption in the log stream.

szkl:
I'll investigate the token verification process through galaxy service next.

apps/passport/app/routes/token.tsx
```diff
@@ -21,33 +23,50 @@ export const action: ActionFunction = async ({ request, context }) => {
...
+      return json(
+        {
+          code: error.data.code,
```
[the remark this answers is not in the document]
szkl: The case here is returning the error message and its code without the details for simplicity. I think the user applications should have a mapping functionality for the backend service errors.

apps/passport/app/routes/token.tsx
szkl: Yeah, `remix-run` logs unhandled errors eventually. Though, any error that doesn't need any functional dependency or representation on the client application shouldn't leave unhandled.

apps/profile/app/utils/session.server.tsx
szkl: Sorry. I described it wrong. The responses with non-200 status codes are not causing an exception.

> A fetch() promise only rejects when a network error is encountered (which is usually when there's a permissions issue or similar). A fetch() promise does not reject on HTTP errors (404, etc.). Instead, a then() handler must check the Response.ok and/or Response.status properties.

[`fetch()`](https://developer.mozilla.org/en-US/docs/Web/API/fetch)

platform/access/src/jsonrpc/methods/exchangeToken.ts
```diff
@@ -235,7 +237,18 @@ const handleRefreshToken: ExchangeTokenMethod<RefreshTokenInput> = async ({
...
+      const message = 'cannot verify token'
```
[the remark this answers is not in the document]
szkl: Yes, most definitely.

szkl:
Pushed a big chunk of changes introducing the `@kubelt/errors` package. Still need to update the pull request description.

apps/passport/app/routes/token.tsx
szkl: This should be only logged and a default error should be thrown instead.

packages/utils/errors.ts
```diff
@@ -0,0 +1,62 @@
...
+      const cause = (error.data as { rollupError: RollupError }).rollupError
+      if (ERROR_CODES[cause.code]) {
+        const ErrorClass = ROLLUP_ERROR_CLASS_BY_CODE[cause.code]
+        return new ErrorClass(cause)
+      } else {
+        return new RollupError(cause)
+      }
+    } else if ('zodError' in error.data) {
+      const cause = (error.data as { zodError: object }).zodError
+      return new Error('zod error', { cause })
```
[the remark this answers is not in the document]
szkl: Updated.

apps/passport/app/utils/authenticate.server.ts
```diff
@@ -40,48 +41,58 @@ export const authenticateAddress = async (
...
+      const authScope = appData.scope
+      const urlParams = new URLSearchParams({
+        client_id: authAppId,
+        redirect_uri: authRedirectUri,
+        state: authState,
+        scope: authScope,
+        prompt: 'none',
+      })
+
+      const grantType = GrantType.AuthenticationCode
```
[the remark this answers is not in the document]
szkl: Good catch.

PR #1803, opened 2023-02-28
feat(profile): improve the connect address flow

### LINKED ISSUE #1702: feat(profile): improve the connect address flow
When we go through the connect address flow we should some more hints along the way:

- [ ] use "connect" query param and "redirect_uri" to point back to our profile app
- [ ] possibly store this query param state in a new cookie `_connect_account_params`
- [x] after authentication when creating the address check if the address is already connected
- [x] if already connected send back with error=ALREADY_CONNECTED query param 
- [x] redirect back using the redirect_uri
- [x] use error query param in profile to explain what happened. if no error show success message.

Cosmin-Parvulescu:
# Description

- Closes #1702 

## Type of change

# How Has This Been Tested?

### DIFF
--- a/apps/passport/app/routes/connect/github/callback.ts
+++ b/apps/passport/app/routes/connect/github/callback.ts
@@ -46,13 +46,19 @@ export const loader: LoaderFunction = async ({
     { alias: profile._json.login, hidden: 'true' }
   )
   const addressClient = getAddressClient(address, context.env)
-  const account = await addressClient.resolveAccount.query({
+  const { accountURN, existing } = await addressClient.resolveAccount.query({
     jwt: await getJWTConditionallyFromSession(request, context.env),
   })
 
   await addressClient.setOAuthData.mutate(authRes)
 
-  return authenticateAddress(address, account, appData, context.env)
+  return authenticateAddress(
+    address,
+    accountURN,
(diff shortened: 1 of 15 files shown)


apps/passport/app/routes/connect/github/callback.ts
```diff
@@ -46,13 +46,19 @@ export const loader: LoaderFunction = async ({
     { alias: profile._json.login, hidden: 'true' }
   )
   const addressClient = getAddressClient(address, context.env)
-  const account = await addressClient.resolveAccount.query({
+  const { accountURN, existing } = await addressClient.resolveAccount.query({
```
maurerbot, line 49: can we not use getAccount?

Cosmin-Parvulescu: I think that resolveAccount was used because this is supposed to be the flow for both fresh / connected flows. So then ultimately the behavior should be that we either create new account and return or return existing account. 

If we use getAccount we would need to handle that case and add code to actually handle account creation. Which I think is a different concern.

maurerbot: ok

maurerbot (approved):

PR #1809, opened 2023-02-28
fix(passport): handle-empty-app-data-in-sign-flow

Cosmin-Parvulescu:
# Description

Error gets thrown if accessing passport directly. This fixes that.

## Type of change

# How Has This Been Tested?

Did the flow with error then without.


poolsar42 (approved):

maurerbot (approved):
shouldn't console validate this?

PR #1817, opened 2023-03-01
chore(passport): connect button show who is connected

### LINKED ISSUE #1773: chore(passport): connect button show who is connected
With wagmi/connect kit we have the ability to get ens profile and connected address. 

We should check for these things and instead of displaying "Login with Wallet" we should say "Login with <ens | address>" and possibly replace the icon with the ens pfp if we have one

Cosmin-Parvulescu:
# Description

Display ensName or address on login button when wallet is connected

- [x] Display connected address
- [x] Display connected ens name
- [x] Display ens avatar

- Closes #1773 

## Type of change

# How Has This Been Tested?

image

image

# Caveats

- There's a flicker between address & ensName if ensName exists. I didn't find any way to interact with the connect button in a way that would allow me to resolve this flicker;
- Displaying ens avatar feels like a large involvement vs. gains for now as the connect button component is very unfriendly with modifying state both inside and outside it.

### DIFF
--- a/apps/passport/app/components/connect-button/ConnectButton.tsx
+++ b/apps/passport/app/components/connect-button/ConnectButton.tsx
@@ -62,8 +62,13 @@ export function ConnectButton({
               >
                 <img src={walletsSvg} />
               </span>
-              {isConnected
-                ? 'Login With Wallet'
+              {isConnected && address
+                ? `Login With ${
+                    ensName ??
+                    `${address.substring(0, 4)} ... ${address.substring(
+                      address.length - 4
+                    )}`
+                  }`
                 : !isConnecting
                 ? 'Connect Wallet'
                 : 'Connecting'}
--- a/apps/passport/app/components/connect-button/ConnectButton.tsx
+++ b/apps/passport/app/components/connect-button/ConnectButton.tsx
@@ -4,6 +4,7 @@ import { Button } from '@kubelt/design-system/src/atoms/buttons/Button'
 import type { ButtonProps } from '@kubelt/design-system/src/atoms/buttons/Button'
 
 import walletsSvg from './wallets.png'
+import { Avatar } from 'connectkit'
 
 const ConnectKitProvider = lazy(() =>
   import('connectkit').then((module) => ({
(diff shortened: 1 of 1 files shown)


apps/passport/app/components/connect-button/ConnectButton.tsx
```diff
@@ -62,8 +62,13 @@ export function ConnectButton({
               >
                 <img src={walletsSvg} />
               </span>
-              {isConnected
-                ? 'Login With Wallet'
+              {isConnected && address
+                ? `Login With ${
+                    ensName ??
```
maurerbot, line 67: I think we also can get their ens pfp?

maurerbot (changes requested):

apps/passport/app/components/connect-button/ConnectButton.tsx
Cosmin-Parvulescu: Hmm, as part of `connectkit`? Was looking at https://docs.family.co/connectkit/api-reference#connectkitbuttoncustom-render-props and there seems to be no reference to ensPfp.

Cosmin-Parvulescu: Ohh, they do seem to have https://docs.family.co/connectkit/api-reference#additional-components-avatar

maurerbot: https://wagmi.sh/react/hooks/useEnsAvatar

Cosmin-Parvulescu: image

apps/passport/app/components/connect-button/ConnectButton.tsx
```diff
@@ -51,19 +60,28 @@ export function ConnectButton({
...
+                    width: 20,
+                    margin: '0 7px',
+                  }}
+                >
+                  <img src={walletsSvg} />
+                </span>
+              )}
+
+              {isConnected && address
+                ? `Login With ${ensName ?? truncatedAddress}`
```
maurerbot, line 84: let's lowercase "with"

maurerbot (changes requested):

apps/passport/app/components/connect-button/ConnectButton.tsx
Cosmin-Parvulescu: `W` => `w`

maurerbot (approved):

PR #1828, opened 2023-03-02
feat(passport): social login not displaying loading bar

### LINKED ISSUE #1782: feat(passport): social login not displaying loading bar
when clicking a social auth button there is some loading time especially if we have already authorized. we should show the loading bar in this case (and all other loading cases).

Cosmin-Parvulescu:
# Description

Added a loading bar when clicking an OAuth login button. This will show a loader while /connect/index flow gets loaded.

- Closes #1782 

## Type of change

# How Has This Been Tested?

Added a 50ms timeout to /github/index.tsx to observe loader in action.

### DIFF
--- a/apps/passport/app/components/connect-oauth-button/index.tsx
+++ b/apps/passport/app/components/connect-oauth-button/index.tsx
@@ -7,8 +7,6 @@ import githubIcon from '@kubelt/design-system/src/assets/social_icons/github.svg
 import googleIcon from '@kubelt/design-system/src/assets/social_icons/google.svg'
 import microsoftIcon from '@kubelt/design-system/src/assets/social_icons/microsoft.svg'
 import twitterIcon from '@kubelt/design-system/src/assets/social_icons/twitter.svg'
-import { useState } from 'react'
-import { Loader } from '@kubelt/design-system/src/molecules/loader/Loader'
 
 type OAuthProvider =
   | 'apple'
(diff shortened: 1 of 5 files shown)


apps/passport/app/components/connect-oauth-button/index.tsx
```diff
@@ -31,22 +33,34 @@ type ConnectOAuthButtonProps = {
...
-          className="w-5 h-5"
-          src={providerIconDict[provider]}
-          alt={provider}
-        />
-      </div>
-    </Button>
-  </form>
-)
+const ConnectOAuthButton = ({ provider }: ConnectOAuthButtonProps) => {
+  const [showLoader, setShowLoader] = useState(false)
```
maurerbot, line 37: is possible to send this state from parent component (maybe the fetcher.state in the route?)

Cosmin-Parvulescu: Hmm, on the route base (`authenticate/index.tsx`) - which calls the `Authentication` component which in turns contains `ConnectOAuthButton` - both `useTransition` as well as `useFetcher` never seem to change their state from:

```
location
: 
undefined
state
: 
"idle"
submission
: 
undefined
type
: 
"idle"
```

Cosmin-Parvulescu: Refactored :))

maurerbot (approved):

PR #1832, opened 2023-03-03
chore(profile): minimizing public profile

### LINKED ISSUE #1808: chore(profile): use outlet context data
# Why

Currently in profile we have the whole profile object in outlet context from the root, however we make separate calls to some profile things in the loaders of root sub-routes. 

We need to not make these additional calls since its unnecessary (and computationally more expensive!) and use data from outlet context instead.

Note: There is an independent `getConnectedAddresses` and a `getAddressProfiles` calls that run sequentially after the main `getProfile` GQL operation.

### LINKED ISSUE #1827: chore(profile): faster profile
### Why

Nav bar context in profile means we do extra slow calls to galaxy to get session context. Cover photo also means during registration we have to add latency to generate gradients.

## What

Simpler profile shape.

- [x] no cover photo in profile (remove from schema too)
- [x] don't generate cover photo in passport
- [x] remove nav bar in public profile
- [x] move session profile fetch to /account root only (delete from root.tsx)
~~- [ ] in public profile only check is user session exsits and if it does get the account urn. If urn matches the profile then show the edit profile button~~
- [ ] use defer if possible to fade in content for illusion of speed
- [x] update styles to match design below
- [x] investigate why ogImage is called twice when navigating from account to my profile
- [x] drop website from profile settings

Image

poolsar42:
### Optimizes public profile

- #### Closes #1827
- #### Closes #1808

# Checklist:


apps/profile/app/routes/account.tsx
```diff
@@ -30,47 +33,80 @@ import {
...
+    } = session.data
+
+    accountURN = parseJwt(jwt).sub as AccountURN
+
+    const fetchedLoggedInProfile = await getAccountProfile({ jwt })
+
+    loggedInUserProfile = fetchedLoggedInProfile
+  }
+
+  if (!loggedInUserProfile) {
```
maurerbot, line 76: this logic should just change to require jwt since there isn't logged in vs logged out in this context

maurerbot (approved):

Issue #1812, opened 2023-02-28
fix(apps): encrypt JWTs in sessin cookies


maurerbot:

szkl:
Option B would be relatively easier to manage and maintain.

szkl:
Option A would be done with an encryption key set in the environment.

PR #1852, opened 2023-03-03
feat(utils): initial `token` utility module

### LINKED ISSUE #1485: chore(passport): copy requireJWT functionality from profile into console and passport

szkl:
# Description

The common operations on the token is defined in a utility module. The functions
in the applications updated to use the functions provided by the utility module.

Closes #1485

## Type of change

# How Has This Been Tested?

- Manual


szkl:
Need to clarify the refresh token capability. https://github.com/proofzero/rollupid/issues/1485#issuecomment-1454267793

szkl:
The expired authentication token causes a redirect to the authentication route.

szkl:
> Just a heads up: In this ticket I talk about maybe storing the access token in a kv and storing the id token int he cookie to match the sub id in kv: #1812
> 
> Would be we able to utilize these utilities (e.g. token expiry) then?

Not like this. The access token expiration time is encoded in the token. The process has to make a request to the backend to get that data.

maurerbot (dismissed):

maurerbot (approved):

PR #1908, opened 2023-03-14
feat(passport): login screen with logged in user

### LINKED ISSUE #1759: feat(passport): Login screen with logged-in user
To enhance the user experience as well as work around issues as noted in #1692, the login screen should be enhanced to show the currently logged in user in addition to the options to log in with another account. This should be the new default behaviour, that replaces the auto-login behaviour that we have today. This is equivalent to `prompt=select_account`.

This new default behaviour should be overridable by the developer to do auto-login by using the `prompt=none` in the `/authorize` call

- [ ] Implement the V1 Figma design in https://www.figma.com/file/EqUEbCGHGnZXDMSUbrxhX7/Kubelt?node-id=8567%3A21023&t=nNneSVPy9uyx4VUJ-0. *Note:* This ticket does NOT implement V2 design.
- [ ] Make `prompt=select_account` and `prompt=none` work as described above.

Image

Cosmin-Parvulescu:
# Description

- Closes #1759

## Type of change

# How Has This Been Tested?

Painfully

### DIFF
--- a/apps/passport/app/routes/connect/apple/callback.ts
+++ b/apps/passport/app/routes/connect/apple/callback.ts
@@ -84,7 +84,7 @@ export const loader: LoaderFunction = async ({ request, context }) => {
     context.traceSpan
   )
   const account = await addressClient.resolveAccount.query({
-    jwt: await getJWTConditionallyFromSession(request, context.env),
+    jwt: await getJWTConditionallyFromSession(request, context.env, appData?.clientId),
     force: !appData || appData.prompt !== 'login',
   })
   const current = await addressClient.getOAuthData.query()
--- a/apps/passport/app/routes/connect/apple/callback.ts
+++ b/apps/passport/app/routes/connect/apple/callback.ts
@@ -85,6 +85,7 @@ export const loader: LoaderFunction = async ({ request, context }) => {
   )
   const account = await addressClient.resolveAccount.query({
     jwt: await getJWTConditionallyFromSession(request, context.env),
+    force: !appData || appData.prompt !== 'login',
   })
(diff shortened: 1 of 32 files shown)


apps/passport/app/routes/connect/apple/callback.ts
```diff
@@ -84,7 +84,8 @@ export const loader: LoaderFunction = async ({ request, context }) => {
     context.traceSpan
   )
   const account = await addressClient.resolveAccount.query({
-    jwt: await getJWTConditionallyFromSession(request, context.env),
+    jwt: await getJWTConditionallyFromSession(request, context.env, appData?.clientId),
+    force: !appData || appData.prompt !== 'login',
```
maurerbot, line 88: what are we doing with this force login?

Cosmin-Parvulescu: `resolveAccount` was hooked for
A - creating new accounts if no jwt present
B - linking to existing account if jwt present

So this force tells `resolveAccount` to created a new account even if jwt is present. `login` is the prompt that gets sent by the account connection flow.

maurerbot: running through the scenarios 
1. jwt present could mean either 
A) we are in a link address flow which is determined by 'login' prompt
B) we have a session (rollup or last) and we are switching to a new one

we want to determin if it's A or B 

2. no jwt present so def a new acct

maurerbot: we will revist this when connect flow in in passport

maurerbot (approved):

PR #1921, opened 2023-03-15
fix(passport): fix passport image panel on small screens

### LINKED ISSUE #1901: fix(passport): fix passport image panel on small screens
Maybe use bg image. solution should allow for develoeprs to set full width/height images in future.

Image

Cosmin-Parvulescu:
# Description

- Closes #1901 

## Type of change

# How Has This Been Tested?

https://user-images.githubusercontent.com/635806/225310459-fc8865a2-199b-49ff-8770-4918a1df7531.mov

### DIFF
--- a/apps/passport/app/components/authentication/Authentication.tsx
+++ b/apps/passport/app/components/authentication/Authentication.tsx
@@ -19,7 +19,7 @@ export function Authentication({
   return (
     <div
       className={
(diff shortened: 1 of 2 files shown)


Cosmin-Parvulescu:
This would be with the background property:
image

Would need to find a way to add the padding from the design if that's a necessity.

maurerbot (approved):

PR #1949, opened 2023-03-20
fix(passport): provision appleid profile

### LINKED ISSUE #1883: bug(passport|profile): apple flow looks successful but not connecting
Repro:
- login to profile
- connect account
- connect apple
- see success message
- no apple account connnected
- replay with same result

szkl:
# Description

The profile data does not include `name` if the user has been authenticated
before.

The id token generation in the access worker requires a `pfp` set.

- Closes #1883

## Type of change

# How Has This Been Tested?

- [x] Manual
  - [x] Connect as the main account
  - [x] Connect as an additional account

### DIFF
--- a/apps/passport/app/utils/authenticate.server.ts
+++ b/apps/passport/app/utils/authenticate.server.ts
@@ -166,9 +166,16 @@ const provisionProfile = async (
             }
           }
           case OAuthAddressType.Apple: {
-            const { firstName, lastName } = res.profile.name!
+            let displayName = ''
+            if (res.profile.name) {
+              const { firstName, lastName } = res.profile.name
+              displayName = `${firstName} ${lastName}`
+            }
             return {
-              displayName: `${firstName} ${lastName}`,
+              displayName,
+              pfp: {
+                image: res.profile.picture,
+              },
             }
           }
           case OAuthAddressType.Discord: {
--- a/apps/passport/app/utils/authenticate.server.ts
+++ b/apps/passport/app/utils/authenticate.server.ts
@@ -166,13 +166,8 @@ const provisionProfile = async (
             }
           }
           case OAuthAddressType.Apple: {
-            let displayName = ''
-            if (res.profile.name) {
-              const { firstName, lastName } = res.profile.name
-              displayName = `${firstName} ${lastName}`
-            }
             return {
-              displayName,
+              displayName: res.profile.name,
               pfp: {
                 image: res.profile.picture,
               },
--- a/platform/access/src/utils/getIdTokenProfileFromAccount.ts
+++ b/platform/access/src/utils/getIdTokenProfileFromAccount.ts
@@ -7,14 +7,13 @@ export default async function (
   ctx: Context
 ): Promise<IdTokenProfile> {
   const accountProfile = await ctx.accountClient.getProfile.query({ account })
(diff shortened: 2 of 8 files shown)


apps/passport/app/utils/authenticate.server.ts
```diff
@@ -166,9 +166,16 @@ const provisionProfile = async (
             }
           }
           case OAuthAddressType.Apple: {
-            const { firstName, lastName } = res.profile.name!
+            let displayName = ''
+            if (res.profile.name) {
```
szkl, line 170: If the user authenticated before then `name` is not present. Therefore need to set a default display name.

apps/passport/app/utils/authenticate.server.ts
```diff
@@ -166,9 +166,16 @@ const provisionProfile = async (
...
+              const { firstName, lastName } = res.profile.name
+              displayName = `${firstName} ${lastName}`
+            }
             return {
-              displayName: `${firstName} ${lastName}`,
+              displayName,
+              pfp: {
+                image: res.profile.picture,
+              },
```
szkl, line 178: [getIdTokenProfileFromAccount()](https://github.com/proofzero/rollupid/blob/303b712f2c80338867d05e85d3530da742a8c9d8/platform/access/src/utils/getIdTokenProfileFromAccount.ts#L10) requires `pfp` to be set.

platform/address/src/jsonrpc/validators/oauth.ts
```diff
@@ -107,12 +107,7 @@ export const TwitterOAuthSchema = z.object({
 export const AppleOAuthSchema = z.object({
   provider: z.literal(OAuthAddressType.Apple),
   email: z.string(),
-  name: z
-    .object({
-      firstName: z.string(),
-      lastName: z.string(),
-    })
-    .optional(),
+  name: z.string().optional(),
```
szkl, line 110: This change is easier than updating the type definition in GraphQL schema.

platform/access/src/utils/getIdTokenProfileFromAccount.ts
```diff
@@ -7,14 +7,13 @@ export default async function (
   ctx: Context
 ): Promise<IdTokenProfile> {
   const accountProfile = await ctx.accountClient.getProfile.query({ account })
-  if (!accountProfile || !accountProfile.displayName || !accountProfile.pfp) {
-    console.error('getIdTokenProfileForAccount', { accountProfile })
+  if (!accountProfile) {
```
szkl, line 10: I think these properties shouldn't break the authorization process.

maurerbot: do we set gradient for apple if not available?

szkl: image

platform/address/src/jsonrpc/validators/oauth.ts
szkl: It is set to an empty string right now. Any defaults?

maurerbot: let's set the name to the email address if empty

maurerbot (approved):

PR #2030, opened 2023-04-04
chore(passport): fetching OTP state with remix fetcher

### LINKED ISSUE #1991: bug(email): Propagate email errors up the call chain
Currently, errors with sending emails (through a `fetch()` against MailChannels) do not propagate up the call chain. This results in a user not being notified that there is an error with sending emails in the Email OTP screen. 

Errors should propagate up, with the correct Error class.
Dependent on #2002 . Test by manually causing a fetch failure in email service's OTP generation code.

### LINKED ISSUE #2002: fix(passport): Handle OTP regen with good ol' `fetch()`
The combination of `state` prop from the `/verify` path, the browser history state, the error catching/rendering is ceasing to be compatible with "the Remix way". Will need to simplify this by resorting to the non-Remix, but Web API-compatible, way of using `fetch()` to call the regen API and handle the error directly in the OTP layout route/component.

All combinations of regen, regen through back-and-forth (with same email), and first-time code entry flows should be tested.

poolsar42:
# Description

Uses fetcher to fetch new state which doesn't update loader history.

- Closes #2002
- Closes #1991 

## Type of change

# How Has This Been Tested?

Locally with the setup

# Checklist:


apps/passport/app/routes/authenticate/$clientId/email/verify.tsx
```diff
@@ -68,24 +73,29 @@ export const action: ActionFunction = async ({ request, context, params }) => {
...
   const ad = useActionData()
-
+  const submit = useSubmit()
   const navigate = useNavigate()
   const transition = useTransition()
+  const location = useLocation()
+  const fetcher = useFetcher()
 
-  const submit = useSubmit()
+  useEffect(() => {
```
maurerbot, line 82: why don't we just grab this from the loader?

maurerbot (changes requested):

apps/passport/app/routes/authenticate/$clientId/email/verify.tsx
```diff
@@ -95,11 +105,13 @@ export default () => {
...
       <EmailOTPValidator
-        loading={transition.state !== 'idle'}
+        loading={transition.state !== 'idle' || fetcher.state !== 'idle'}
         email={address}
-        state={matchData?.state}
+        state={fetcher.data?.state}
         invalid={ad?.error}
-        requestRegeneration={() => navigate(`?address=${address}`)}
+        requestRegeneration={async () => {
+          fetcher.load(`/connect/email/otp${location.search}`)
```
maurerbot, line 113: use `fetch` not fetcher and catch the 400 and use `setErrorState`.

Remember to `setErrorState` to null when fetching this

apps/passport/app/routes/connect/email/otp.ts
```diff
@@ -32,7 +32,7 @@ export const loader: LoaderFunction = async ({ request, context }) => {
     return json({ state })
   } catch (e) {
     console.error('Error generating email OTP', e)
-    throwJSONError(e)
```
maurerbot, line 35: keep this

apps/passport/app/routes/authenticate/$clientId/email/verify.tsx
maurerbot: when you call the loader wrap it in a try / catch and set in `useLoaderData` to check for error.

If error `setErrorState(error)`

maurerbot (changes requested):

maurerbot (changes requested):

poolsar42:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/toString

apps/passport/app/routes/authenticate/$clientId/email/verify.tsx
poolsar42: Not sure what you meant here, but here's what I [did](https://github.com/proofzero/rollupid/pull/2030/files#diff-68d6f036dae98e910be2f10f435d39df7de03cc3b3a34f6cf570a20f979ed2cbR96)

It was based on [Error.prototype.toString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/toString) function

maurerbot (approved):

PR #2056, opened 2023-04-10
chore(passport): otp time update

### LINKED ISSUE #1975: chore(passport): Delete unused OTP codes and address on expiry
When a new code is generated, the state value of the old one isn't preserved on the caller side. Since state is a required parameter for verification, without it the old codes can't be verified. As such, the code to maintain multiple codes should be removed.

Also, when the code alarm is fired, the address node should delete the DO (invoking `deleteAll()` on its storage object)

### LINKED ISSUE #2014: bug(passport): cancel otp does not delete current otp state
~~if you cancel email login and try again you get the error that you can't send another code.~~

~~if you cancel we should invalidate the old code and start from scratch.~~

Update: works as designed, but the design needs to change for better UX. Decision: no change in how the delay is applied, however, delay between attempts should be lowered to 30 sec, complemented by an upper limit of 5 attempts within 5 minutes, that, once triggered, expires after 10 minutes.

poolsar42:
# Description

Introduces new time limits for OTP sends:
- [x] 5 OTPs might be send only within 5 minutes period. If this limit is heat - user'll wait 10 minutes to be able to get another one
- [x] 30 seconds between consequent OTP requests

- Closes #2014
- Closes #1975

## Type of change

# How Has This Been Tested?

To test error propagation and  error messages for the different limits I set time frames to lower limits (5 seconds instead of 30 and 30 seconds instead of 10 minutes) in [this file](https://github.com/proofzero/rollupid/pull/2056/files#diff-58cbdbd3330f18374ae50b879d52c36164f9df75a5b2253a1ac143971817cb35) and then was testing it in passport with "connect with email" flow

# Checklist:


platform/address/src/nodes/email.ts
```diff
@@ -27,16 +31,47 @@ export default class EmailAddress {
...
+        throw new BadRequestError({
+          message: `Cannot generate new code for the address.
+          You can only generate a new code 5 times every ${EMAIL_VERIFICATION_OPTIONS.regenDelayFor5SubsCallsInMins}
+          minutes`,
+        })
+      }
+
+      // Every 10 minutes we wipe the number of attempts and first attempt timestamp
+      if (
+        payload.record.firstAttemptTimestamp +
```
maurerbot, line 69: why not set an alarm and wipe the data with a timer

platform/address/src/nodes/email.ts
```diff
@@ -27,16 +31,47 @@ export default class EmailAddress {
         CODES_KEY_NAME
```
maurerbot, line 31: why aren't we just storing the payload under `codes/{state}`. We wound't have to iterate over every single one

platform/address/src/nodes/email.ts
```diff
@@ -27,16 +31,47 @@ export default class EmailAddress {
...
       )) || {}
 
+    let numberOfAttempts
+    let firstAttemptTimestamp
+    let keepAttemptsNumber = true
+
     for (const [_, payload] of Object.entries(verificationCodes)) {
       if (
-        payload.creationTimestamp + EMAIL_VERIFICATION_OPTIONS.regenDelayInMs >
+        payload.record.creationTimestamp +
```
maurerbot, line 40: we aren't even filtering by state?

maurerbot (changes requested):

platform/address/src/nodes/email.ts
maurerbot: so, `codes/{code}` and `code/*` to list in alarm

for example: 

`this.storage.list("codes/")`

maurerbot: https://developers.cloudflare.com/workers/runtime-apis/durable-objects/#transactional-storage-api

maurerbot (changes requested):
Changes:
- single code field which stores increments and times
- if incrementing / issuing code set 5 min alarm
- if max attempts set alarm to 10 min
- alarm should check for node.getAccount. If null deleteAll otherwise delete code field

platform/address/src/nodes/email.ts
```diff
@@ -27,35 +31,96 @@ export default class EmailAddress {
...
+            Math.floor((timeLeft / 1000) % 60)
+              ? `and ${Math.floor((timeLeft / 1000) % 60)}  seconds`
+              : ''
+          }`,
+        })
+      }
+
+      // We count 5 minutes from the first attempt
+      if (
+        verificationPayload.numberOfAttempts >
```
maurerbot, line 74: shoult this be `===` max attempts?

platform/address/src/nodes/email.ts
```diff
@@ -27,35 +31,96 @@ export default class EmailAddress {
...
+              60
+            } minutes.`,
+        })
+      }
     }
 
     const creationTimestamp = Date.now()
+    let numberOfAttempts = 1,
+      firstAttemptTimestamp = creationTimestamp
+    if (verificationPayload) {
```
maurerbot, line 106: why have this condition again if you can do it in the large if statement above (line 37). Then I'd actually set the check for max attempts here and set the alarm if attempts reached. that way in the condition above you don't set the alarm and it's just an early termination. Right now the alarm can be pushed again if that condition is reached around line 100 and return

platform/address/src/nodes/email.ts
```diff
@@ -82,34 +156,47 @@ export default class EmailAddress {
...
-      (await address.state.storage.get<Record<string, VerificationPayload>>(
-        CODES_KEY_NAME
-      )) || {}
+    const verificationPayload =
+      await address.state.storage.get<VerificationPayload>(OTP_KEY_NAME)
 
-    for (const [code, verification] of Object.entries(codes)) {
-      if (
-        verification.creationTimestamp + EMAIL_VERIFICATION_OPTIONS.ttlInMs <=
+    if (
```
maurerbot, line 173: you can probably be confident that if the alarm is run you don't need to verify anything. Just check if there is an account and delete the code otherwise deleteAll. Technically deleteAll isn't needed because if we delete the code the and there is nothing else stored the DO will be garbage collected anyway.

maurerbot (changes requested):

platform/address/src/nodes/email.ts
poolsar42: I keep >= in case something went over the 5

platform/address/src/nodes/email.ts
poolsar42: > you can probably be confident that if the alarm is run you don't need to verify anything. Just check if there is an account and delete the code otherwise deleteAll. Technically deleteAll isn't needed because if we delete the code the and there is nothing else stored the DO will be garbage collected anyway.

I think we store provided email address as an `alias` field. In this case it doesn't matter I guess if we call `.deleteAll()` or `.delete('alias')`

platform/address/src/nodes/email.ts
```diff
@@ -27,51 +31,127 @@ export default class EmailAddress {
...
-    await this.node.storage.put(CODES_KEY_NAME, verificationCodes)
-    await this.node.storage.setAlarm(
-      creationTimestamp + EMAIL_VERIFICATION_OPTIONS.ttlInMs
-    )
+    await this.node.storage.put(OTP_KEY_NAME, payload)
+
+    // We have 2 types of alarms to set
+    // 1. If the user has hit the limit of 5 attempts, we set a cool down of 10 minutes
+    if (cooldownStartTimestamp) {
+      await this.node.storage.setAlarm(
```
maurerbot, line 125: set this alarm in the `if (verificationPayload)` block

platform/address/src/nodes/email.ts
```diff
@@ -27,51 +31,116 @@ export default class EmailAddress {
...
+      firstAttemptTimestamp = verificationPayload.firstAttemptTimestamp
+      // we limit the number of attempts to 5 within 5 minutes
+      // once the limit of 5 attempts is hit, we set a cool down 10 minutes from the last attempt
+      if (
+        numberOfAttempts >= EMAIL_VERIFICATION_OPTIONS.maxAttempts &&
+        firstAttemptTimestamp +
+          EMAIL_VERIFICATION_OPTIONS.maxAttemptsTimePeriod >
+          Date.now()
+      ) {
+        cooldownStartTimestamp = creationTimestamp
```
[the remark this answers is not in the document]
maurerbot: you don't even need this if you just set the 10 min alarm if this checks out. That's why the condition should be if `numberofAttempts + 1 === options.maxAttempts`. Then you store `numberofAttempts + 1` and set the alarm and you'll never hit the alarm condition again.

platform/address/src/nodes/email.ts
maurerbot: Ok. but you still don't need to check the condition. Only run the account condition for delete vs delete all.

maurerbot (changes requested):

maurerbot (approved):

maurerbot (dismissed):

PR #2083, opened 2023-04-13
bug(passport): Connected account toast and button after cancelation

Cosmin-Parvulescu:
### Description

* Hacky
- Added a connected account count to the connection_requested cookie
- If screen reload is not more than set count, it means no connection was made so we don't display success
- Alternative is to add a query param for success that we can interrogate
- Added hard coded cancellation url to back button in authentication screen
- This clears console params

### Related Issues

- Closes #2071 

### Testing

As described in ticket

### Checklist

- [ ] I have read the [CONTRIBUTING](CONTRIBUTING.md) guidelines
- [ ] I have tested my code


maurerbot (changes requested):
Didn't this used to do it?  https://github.com/proofzero/rollupid/pull/2079/files#diff-6634693d9b7bae7961105dddb7e4359569a160c35196b8176c3cd6600b9d05d0L44

PR #2093, opened 2023-04-13
fix(address): keep the stored refresh token if not provided

szkl:
---

### Description

The object in the response if does not have a refresh token then the property in
the stored object is removed.

This change puts the stored refresh token into the object if there is none in
the response.

### Related Issues

- Closes #2089

### Testing

- [x] Manual

### Checklist

- [ ] I have tested my code
- [x] I have updated the documentation (if necessary)


szkl:
The test is going to take some time.

PR #2097, opened 2023-04-13
fix(passport): redirect uri validation

### LINKED ISSUE #2060: bug(passport): Implement Redirect URI validation
Redirect URI coming from the authorization request should be validated against the registered redirect URI for the application/client. The origin should match, but the path can be different.

poolsar42:
### Description

Checks redirectURI on authorize screen with the one configured in console.

### Related Issues

- Closes #2060

### Testing

1. went to authorize screen with invalid redirect URL (different from the one that was set in console)
2. got an error
3. changed it so redirect URI in link matches with one in console
4. worked
5. went back to invalid redirect URI, updated it in console configurations so they match
6. reloaded page -> and it worked

### Checklist

- [x] I have tested my code

### DIFF
--- a/apps/passport/app/routes/authorize.tsx
+++ b/apps/passport/app/routes/authorize.tsx
@@ -156,9 +156,18 @@ export const loader: LoaderFunction = async ({ request, context }) => {
       }),
     ])
 
+    const providedUrl = new URL(redirectUri)
+    const configuredUrl = new URL(appPublicProps.redirectURI)
+
     if (!appPublicProps.scopes || !appPublicProps.scopes.length)
       throw new BadRequestError({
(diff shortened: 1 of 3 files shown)


maurerbot (dismissed):

apps/passport/app/routes/authorize.tsx
```diff
@@ -156,11 +156,20 @@ export const loader: LoaderFunction = async ({ request, context }) => {
       }),
     ])
 
+    const providedUrl = new URL(redirectUri)
```
maurerbot, line 159: @betimshahini if provided URL is null do we want to set default from appData?

maurerbot (approved):

PR #2125, opened 2023-04-18
feat(platform): Connected address scope with no user-selectability

### LINKED ISSUE #2023: feat(passport): Connected address scope without user-selectability
Follow on from #1990. Enhance the authorization functionality to support the scope for selectable connected addresses.

- [x] Persona structure needs to be enhanced with the new scope. Assume an array of addressURNs as the value of the record coming back from authorization screen.
- [x] Scope should be available for selection in Console OAuth page
- [x] Scope should be available in text form in authz screen, when included in the `scope` param of the request and is configured as an allowed scope value in Console
- [x] Access/refresh tokens should contain the new scope value in `scope` field, when authorized
- [x] ID token should also include the result of the connected addresses structure.
- [x] Userinfo endpoint should show the structure of authorized connected addresses:

Example `personaData` data structure:
```
personaData = {
  // ... other elements
  connected_addresses: [
    'selectedAddressUrn1',
    'selectedAddressUrn2',
  ]
}
```
Example `/userinfo` and ID token data structure:
```
connected_addresses: [
  {
    type: 'github',
    alias: 'somegithubusername'
  },
  {
    type: 'google',
    alias: 'someemail@gmail.com'
  },
  {
    type: 'eth',
    alias: '0x123456789012345678901234567890123456789012'
  }
]
```

### LINKED ISSUE #2099: feat(passport): Add support for `prompt=consent` param in  `/authorization`
# Why

With the introduction of skip authz if already authz developers may want to force reauth if they lose something like a refresh token.

# What

Introduce the `prompt=consent` query param

# How
 
Skip pre-authz when `prompt=consent` is in the authz query params

betimshahini:
### Description

Implements
- [x] Connected address scope in Console OAuth page
- [x] Connected address in authz screen when requested as part of `scope` param of `/authorize`
- [x] `connected_account` in access token `scope` field
- [x] `connected_account` list in ID token and `/userinfo` reponse.
- [x] `prompt=consent` support for `/authorize` endpoint

### Related Issues

- Closes #2023
- Closes #2099

### Testing

- Added `connected_addresses` as part of "Allowed scope" field in Console for a test app.
- For the same app, crafted a `/authorize` request with `scope=email+connected_addresses` which were presented on the authz screen. After authorization, exchanged the returned code for tokens and verified ID token, access token and `/userinfo` response using the access token.
- After authorizing the request once, add `prompt=constent` to the authz request URL so that you get re-prompted to authorize.
 
### Checklist

- [x] I have tested my code
- [x] I have updated the documentation (if necessary)


packages/security/persona.ts
```diff
@@ -55,6 +65,41 @@ export async function validatePersonaData(
...
+      ) {
+        throw new BadRequestError({
+          message: 'Bad data received for list of address identifiers',
+        })
+      }
+
+      const accountClient = createAccountClient(env.accountFetcher, {
+        ...generateTraceContextHeaders(traceSpan),
+      })
+      const accountAddresses = await accountClient.getAddresses.query({
```
maurerbot, line 87: can the query run filter?

packages/security/persona.ts
```diff
@@ -70,28 +115,67 @@ export async function setPersonaReferences(
...
+  //TODO: The next set of 3 operations will need to be optmizied into a single
+  //SQL transaction
+
+  //Get existing references
+  const edgesToDelete = await edgesClient.getEdges.query({
+    query: { tag: EDGE_AUTHORIZES_REF_TO, src: { baseUrn: accessNode } },
+  })
+
+  //Delete existing references
+  edgesToDelete.edges.forEach(
```
maurerbot, line 151: why delete?

packages/security/persona.ts
```diff
@@ -150,6 +234,27 @@ export async function getClaimValues(
           picture: nodeResult.qc.picture,
         }
       }
+    } else if (scopeValue === 'connected_addresses') {
```
maurerbot, line 237: we need to start optimizing this. I feel like these could be trpc calls to the address or account worker. Maybe something like [selective disclosures](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-04.html)?

Generally, the pattern we want is the specific service does it's own verification and the access node defers to that service to verify the claims. The verifier should be encoded in the scope values, that is the intention of the url based scopes.

packages/security/persona.ts
maurerbot: Are we creating edges for each authorized address? I thought each authorization is a new access token. The only edges we need to maintain are the edges between the account<->access node->app and the access node stores all the unique sets of claims for access token validation.

packages/security/persona.ts
maurerbot: We can differentiate the display scope value vs the actual scope value in the security package (and if supporting third parties in the future probably a registration service.  Example: https://developers.google.com/oauthplayground/

I think the validation that the scope claims are valid can be handled by the access worker before reaching out to the delegated worker to deal with the results. Also noting that this stuff should probably not be in packages/security -- packages should not deal with side-effects (e.g. creating and calling clients).

packages/security/persona.ts
maurerbot: for only the addresses in the claims so you don't need to run like 93

packages/security/persona.ts
maurerbot: Why does this have to be edges at all? Seems like overkill if we have the access node storing similar if not more information?

maurerbot: In the case of "deleting identity" and cascading authz check, that could be done on a per authz node basis to prompt the user to first revoke access to all their apps, membership of groups, etc. Each claim being an edge doesn't make this easier

maurerbot (approved):
recap on convo:
- In the optimization issue we will address side-effects in packagers and instead use a dependency injection pattern
- scopes use colon delimitation but we need to work towards value registration and retrieval to be general enough noting that with service bindings we will have to deal with our first party scopes until (maybe) moving away from service bindings toward a workers for platform model.

PR #2170, opened 2023-04-25
fix(passport): If name of email is the same for two types of email addresses they both look selected

### LINKED ISSUE #2107: fix(passport): If name of email is the same for two types of email addresses they both look selected
### Description

Using name instead of id to show checkmark on select creates edge cases as seen in screenshot

### Steps to Reproduce

1. Connect a SSO email
2. Connect an Email using the same as one of the SSO
3. Login with email scope

### Expected Behavior

Only show checkbox for actual selected

### Actual Behavior

Multiple look selectd

### Screenshots (if applicable)
image

Cosmin-Parvulescu:
### Description

- [x] Updated email-select item identifier from `email` to `id` (address urn)

### Related Issues

- Closes #2107 

### Testing

- [x] Added duplicate e-mail in storybook props
- [x] Reproduced condition in storybook
- [x] Updated item identifier
- [x] Condition no longer reproduces

### Checklist

### DIFF
--- a/packages/design-system/src/atoms/email/EmailSelect.tsx
+++ b/packages/design-system/src/atoms/email/EmailSelect.tsx
@@ -71,7 +71,7 @@ export const EmailSelect = ({
       onChange={(selected) => {
         setSelected(selected)
       }}
-      by="email"
+      by="addressURN"
     >
       {({ open }) => (
         <div className="relative bg-white">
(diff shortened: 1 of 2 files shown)


packages/design-system/src/atoms/email/EmailSelect.tsx
```diff
@@ -71,7 +71,7 @@ export const EmailSelect = ({
       onChange={(selected) => {
         setSelected(selected)
       }}
-      by="email"
+      by="addressURN"
```
maurerbot, line 74: doe this change what is submitted in the form?

Cosmin-Parvulescu: It shouldn't. Form submission gets populated by the entire list item and selected property:
`personaData.email = selectedEmail.addressURN` which is further validated by:
```
if (scopeName === 'email') {
      const addressUrnForEmail = claimValue
      if (!AddressURNSpace.is(addressUrnForEmail))
        throw new BadRequestError({
          message: 'Bad data received for address identifier',
        })
        ...
```

maurerbot (approved):

PR #2208, opened 2023-05-02
feat(passport): prevent identity deletion when it owns an app

poolsar42:
### Description

Prevents deletion of identity when there's an app attached to the identity

### Related Issues

- Closes #2148 

### Testing

2 cases were tested:
- [ ] create an app and try to delete an identity that has an app associated with it - the modal prevents me to do so
- [ ] easily deletes an identity that doesn't have any apps associated with it

### Checklist


szkl (dismissed):

apps/passport/app/routes/settings.tsx
```diff
@@ -101,6 +103,7 @@ export const loader: LoaderFunction = async ({ request, context }) => {
     connectedProfiles: normalizedConnectedProfiles,
     CONSOLE_URL: context.env.CONSOLE_APP_URL,
     primaryAddressURN: accountProfile?.primaryAddressURN,
+    ownedApps,
```
maurerbot, line 106: why don't we do these checks when we press the button. seems too eager to do it ahead of time.

maurerbot (changes requested):

apps/passport/app/routes/settings.tsx
poolsar42: Since we fetch everything for all routes under `settings` in passport, I decided to follow the same pattern here.

More to that, in order to fetch `ownedApps` from internal  `settings/advanced`route we'll have to validate jwt in here, but at the same time we'll do jwt prepwork in the parental `settings` route (if I'm not mistaken remix calls all loaders for the route)

maurerbot: I see. Do we validate again after?

poolsar42: Nope, [here's ](https://github.com/proofzero/rollupid/pull/2208/files#diff-db36cde6843ac3bcf12a68dbde019433ce40871151c04e1790bf517f372ae1caR39) the only place where we do it and then [pass](https://github.com/proofzero/rollupid/pull/2208/files#diff-db36cde6843ac3bcf12a68dbde019433ce40871151c04e1790bf517f372ae1caR47) it to starbase client and then [listApps](https://github.com/proofzero/rollupid/pull/2208/files#diff-db36cde6843ac3bcf12a68dbde019433ce40871151c04e1790bf517f372ae1caR87) with it. In list apps we validate it again

maurerbot: Validating user owns app on load and then not again when we trigger that user deleting is no good

maurerbot (changes requested):

PR #2236, opened 2023-05-04
feat(passport): Revocation of smart contract wallet scope

### LINKED ISSUE #2119: feat(passport): Revocation of smart contract wallet scope
Dependent on #2117, #2118 

When a user initiates a revocation of the authorization against an application that has the `smart_contract_wallet` scope, in addition to the revocation functionality that is done today (clearing out the access DO, removing node references from edges, etc), the revocation process for smart contract wallets should update the smart contract with immediate effect to invalidate the session keys generated for that app/developer.

Session keypair's public key will be stored in a new structure in the Access DO called `appData` at the point the key is registered.

Note: that this will be optimized further in #2289

### LINKED ISSUE #2201: chore(integration): Test validity of 4337 session key and update docs
Test validity of 4337 session key with gas-only transaction (no currency) as well as a currency-based transaction. Do this on Goerli.

Based on above, update docs with quick tutorial on how to use,

### LINKED ISSUE #2410: fix(passport): revoke button in passport doesn't disable while submitting

poolsar42:
### Description

Revoke Smart contract wallet session key

Updates docs to indicate revocation of SC wallet session keys

### Related Issues

- Closes #2119 
- Closes #2410
- Closes #2201

### Testing

1. Authorise app for erc_4337 scope: 
`localhost:10001/authorize?client_id=b84308327fccec7d52d454011286c6d8&redirect_uri=http://localhost&state=foo&scope=erc_4337&prompt=consent`
2. After it's done, query galaxy to create session key for this SC wallet:
`curl http://localhost:10401/graphql \
  -X POST \
  -H 'content-type: application/json' \
  -H 'X-GALAXY-KEY: %GALAXY APY KEY%' \
   -H 'Authorization: Bearer %ACCESS TOKEN%' \
  --data '{"query":"mutation { registerSessionKey (sessionPublicKey: \"%DEVELOPER'S WALLET ADDRESS%\", smartContractWalletAddress: \"%SMART WALLET BLOCKCHAIN ADDRESS%\" )}"}'`

This endpoint will return session key, test then this session key - e.g. mint NFT with this. Here' my [script](https://gist.github.com/poolsar42/ea3f671fccda9f51c6edbb16146d23b1)

3. Revoke Access for authorised app
4. Try minting NFT with same session key once again

### Checklist


apps/passport/app/components/applications/claims.tsx
```diff
@@ -1,4 +1,4 @@
-import { useFetcher } from '@remix-run/react'
```
poolsar42, line 1: This is the only valuable change in this file - I change `fetcher.Form` on `Form` from `@remix-run/react`. Other changes here are from prettier.

platform/address/src/context.ts
```diff
@@ -43,6 +43,7 @@ interface CreateInnerContextOptions
   DELAY_BETWEEN_REGENERATION_ATTEMPTS_IN_MS: number
   REGENERATION_COOLDOWN_PERIOD_IN_MS: number
   MAX_ATTEMPTS_TIME_PERIOD_IN_MS: number
+  MUMBAI_PROVIDER_URL: string
```
poolsar42, line 46: Keep it as Mumbai for now - will switch to ethereum mainnet if needed

platform/address/src/types.ts
```diff
@@ -25,6 +25,7 @@ export interface Environment {
   DELAY_BETWEEN_REGENERATION_ATTEMPTS_IN_MS: number
   REGENERATION_COOLDOWN_PERIOD_IN_MS: number
   MAX_ATTEMPTS_TIME_PERIOD_IN_MS: number
+  MUMBAI_PROVIDER_URL: string
```
poolsar42, line 28: Same about Mumbai in different file

platform/address/src/context.ts
maurerbot: should be based on project id?

maurerbot (changes requested):

platform/address/src/context.ts
poolsar42: should be based on the env I'd say and project id should be based on env as well (I think this is how it is now, right?)

maurerbot: what env? project id's are tied to the chain.

platform/access/src/jsonrpc/methods/revokeAppAuthorization.ts
```diff
@@ -94,5 +108,55 @@ export const revokeAppAuthorizationMethod: RevokeAppAuthorizationMethod =
...
+    })
+    const paymaster = await starbaseClient.getPaymaster.query({
+      clientId,
+    })
+
+    const appData = await caller.getAppData({
+      clientId,
+    })
+
+    if (appData?.smartWalletSessionKeys?.length) {
```
maurerbot, line 136: this object should have the project id that was used to register the session key? It should also check if the session key is expired or not so we don't have to revoke it?

platform/address/src/context.ts
poolsar42: I am talking about our env - for everything not in prod it should probably be tied to Mumbai and only mainnet on prod

platform/access/src/jsonrpc/methods/revokeAppAuthorization.ts
maurerbot: to be more generic, it should state the provider and the the provider secret used to register it. 

e.g. `provider: ZeroDev, config: {projectId: <project id>}`

poolsar42: This revocation happens only for a specific app. In our current schema we have `getPaymaster` endpoint in starbase which returns exactly this generic information about paymaster provider you mentioned in the last comment.

```ts
export const PaymasterProviderSchema = z.literal('zerodev')

export const PaymasterSchema = z
  .object({
    provider: PaymasterProviderSchema,
    secret: z.string(),
  })
  .optional()
```

as for `appData` it stores smart wallet addresses and dev's address to which session key was issued. Since per one app we have one session key there's no point of storing it additionally in `appData`.

maurerbot (approved):

PR #2248, opened 2023-05-08
fix(passport): infinite loop in account select

Cosmin-Parvulescu:
### Description

- [x] Fixed infinite loop in account select
- [x] Fixed infinite errors in remix apps due to unset debug port

### Testing

- [x] Checked console log for debug errors
- [x] Used console log to display selected accounts to make sure it still works as intended 

### Checklist

### DIFF
--- a/apps/console/app/root.tsx
+++ b/apps/console/app/root.tsx
@@ -85,12 +85,6 @@ export type LoaderData = {
 }
 
 export const loader: LoaderFunction = async ({ request, context }) => {
-  if (!process.env.REMIX_DEV_SERVER_WS_PORT) {
-    throw new InternalServerError({
-      message: 'REMIX_DEV_SERVER_WS_PORT is not defined',
-    })
-  }
-
   const jwt = await requireJWT(request)
   const traceHeader = generateTraceContextHeaders(context.traceSpan)
   const parsedJwt = parseJwt(jwt)
@@ -136,7 +130,7 @@ export const loader: LoaderFunction = async ({ request, context }) => {
         INTERNAL_GOOGLE_ANALYTICS_TAG,
         REMIX_DEV_SERVER_WS_PORT:
           process.env.NODE_ENV === 'development'
-            ? +process.env.REMIX_DEV_SERVER_WS_PORT
+            ? +process.env.REMIX_DEV_SERVER_WS_PORT!
             : undefined,
       },
       displayName,
--- a/apps/console/app/root.tsx
+++ b/apps/console/app/root.tsx
@@ -191,7 +191,7 @@ export default function App() {
           }}
         />
         <Scripts nonce={nonce} />
-        <LiveReload nonce={nonce} />
+        <LiveReload nonce={nonce} port={8002} />
       </body>
     </html>
   )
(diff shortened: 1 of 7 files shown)


apps/console/app/root.tsx
```diff
@@ -191,7 +191,7 @@ export default function App() {
           }}
         />
         <Scripts nonce={nonce} />
-        <LiveReload nonce={nonce} />
+        <LiveReload nonce={nonce} port={8002} />
```
szkl, line 194: Can you try getting it from `[process.env.REMIX_DEV_SERVER_WS_PORT]`(https://github.com/remix-run/remix/blob/bac8b5bf9c9c483a6aaed2db813782a546e40148/packages/remix-dev/config.ts#LL669C3-L669C39)

apps/console/app/root.tsx
```diff
@@ -77,10 +80,17 @@ export type LoaderData = {
   displayName: string
   ENV: {
     INTERNAL_GOOGLE_ANALYTICS_TAG: string
+    REMIX_DEV_SERVER_WS_PORT?: number
   }
 }
 
 export const loader: LoaderFunction = async ({ request, context }) => {
+  if (!process.env.REMIX_DEV_SERVER_WS_PORT) {
+    throw new InternalServerError({
```
szkl, line 89: Shouldn’t throw in production build.

packages/design-system/src/hooks/useTreeshakeHack.tsx
```diff
@@ -0,0 +1,19 @@
...
+ * unused code from the final bundle. By logging the value to the console, it creates
+ * a side-effect that the bundler cannot safely remove, thus preventing it from being
+ * tree-shaken.
+ **/
+export default function <T>(value: T) {
+  useEffect(() => {
+    if (value) {
+      console.log(`Prevented ${value} from getting tree shaken`)
+    }
+  }, [])
```
szkl, line 18: value should be in the dependency list

apps/console/app/root.tsx
Cosmin-Parvulescu: Probably shouldn't throw at all.

packages/design-system/src/hooks/useTreeshakeHack.tsx
Cosmin-Parvulescu: Wouldn't matter in this scenario; added it to address the concern. :)

packages/design-system/src/atoms/accounts/ConnectedAccountSelect.tsx
```diff
@@ -43,10 +45,17 @@ export const ConnectedAccountSelect = ({
...
         onSelectAll()
       }
-    } else {
-      if (onSelect) {
-        onSelect(selectedAccounts)
-      }
+    }
+  }, [allConnectedAccountsSelected])
+
+  useEffect(() => {
```
maurerbot, line 51: isn't this a good useCase for useMemo?

https://stackoverflow.com/questions/56028913/usememo-vs-useeffect-usestate

Cosmin-Parvulescu: AI Overlord:

> `useMemo` and `useEffect` serve different purposes, so they are not interchangeable. Keep in mind that `useMemo` is specifically designed for memoizing the result of a function to avoid unnecessary computations, while `useEffect` is designed to handle side effects, like data fetching or DOM manipulation.
> 
> In our case, it's not possible to replace `useEffect` with `useMemo` without changing the behavior of the code. Both `useEffect` hooks are dealing with side effects (updating the state and calling functions with side effects) that should not be handled by `useMemo`.

There's probably a better way to handle these kind of situations, but I would suggest we spend the time to find it when we create the generic select component because ideally we'll end up replacing all current implementations with something based on that.

maurerbot (approved):

PR #2262, opened 2023-05-09
feat(passport): generic scope for system identifiers

poolsar42:
### Description

Creates a "System identifiers" generic scope for scopes like `openid` which don't require personal information, but for user to be concerned about what they share

### Related Issues

- Closes #2090

### Testing

`openid` scope in `/authorize` route

### Checklist


maurerbot (changes requested):
system identifier is not a scope
CleanShot 2023-05-10 at 08 16 33@2x

maurerbot (dismissed):

PR #2427, opened 2023-06-16
fix(console): remove aspect ratios in designer

### LINKED ISSUE #2408: fix(console): remove aspect ratios in designer
Restricting aspect ratio makes finding the right image hard. it also doesn't full solve the issues (see image below). We should just use css cover and let users upload w/e they like. 

For email 1:1 is oncorrect anyway.. They can preview the email and choose to upload w/e size they want.

Image

poolsar42:
### Description

Remove aspect ratios for designer

### Related Issues

- Closes #2408

### Testing

https://github.com/proofzero/rollupid/assets/33105890/6b2c8ca9-5a88-4965-ba22-eb0080998095

### Checklist

### DIFF
--- a/apps/console/app/routes/apps/$clientId/designer.beta.tsx
+++ b/apps/console/app/routes/apps/$clientId/designer.beta.tsx
@@ -597,17 +597,11 @@ const AuthPanel = ({
 
           <FormElement
             label="Login Screen Side Image"
-            sublabel="2:3 ratio (at least 720x1080px), images can't be larger than 2mB"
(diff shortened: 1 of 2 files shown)


poolsar42:
> Do we also need to change this to a css cover? rather than image tag?

already there https://github.com/proofzero/rollupid/blob/main/apps/passport/app/routes/authenticate/%24clientId.tsx#L102

poolsar42:
> this ticket also needs the change in email template.

Understood. Now it keeps track only of max-width and changes ratios only in case when width of provided image is larger

maurerbot (approved):

PR #2457, opened 2023-07-03
feat(console): Billing and Invoicing - Invoice and payments integration

### LINKED ISSUE #2404: feat(console): Billing and Invoicing - Invoice and payments integration
Depends on #2389.
Context for Stripe API: https://github.com/proofzero/rollupid/issues/2334#issuecomment-1587404215

Implement integration of invoice table and Update payment information button:
- [x] Invoice information should be retrieved using Customer ID stored in user's Account DO, under `servicePlan` key.
- [x] Button should redirect to Stripe customer portal with Customer ID
- [x] A reconciliation function should be built to be used to check Stripe subscriptions SKUs (amounts and prices), and match them against the same in `servicePlan`. Any deltas should be updated in `servicePlan`. Furthermore, any removal of any application-assigned paid plans should be reflected as unassignments to those apps, in the order of most recent edges between AccountURN and ApplicationURN, with tag `pays/app`.
- [x] Any unassignments from reconciliation method above should be followed by an email notification to the billing notification email (to notify of the billing change) and the developer notification emails for each app affected (to notify of the removal of the paid features)
- [x] Webhooks for Stripe subscription changes should be built covering updates, pauses, resumes and deletes. The webhooks should use the reconciliation method above to process the requests.

Cosmin-Parvulescu:
### Description

- Added Invoices to billing screen
- Added reconciliation function for automagically downgrading apps if entitlements get removed through Stripe subscription
- Added two new e-mail templates for reconciliation (billing + dev)
- Added Portal endpoint for redirecting users to their Stripe Customer Portal
- Parametrized getEntitlements to be able to be used from public webhook
- 

### Related Issues

- Closes #2404 
- Pause / unpause to be handled at a later date

### Testing

- Did some small local testing, would need to do additional testing in dev

### Checklist

### DIFF
--- a/apps/console/app/routes/__layout/gnillib/index.tsx
+++ b/apps/console/app/routes/__layout/gnillib/index.tsx
@@ -1104,112 +1104,175 @@ export default () => {
             </Link>
           </header>
 
-          <main className="flex flex-row gap-4 items-center border rounded-lg">
-            <table className="min-w-full bg-white rounded-lg overflow-hidden">
-              <thead className="bg-gray-50 rounded-t-lg">
-                <tr className="rounded-t-lg">
-                  <th className="px-6 py-3 text-left rounded-tl-lg">
-                    <button
-                      className="flex flex-row gap-2"
-                      onClick={() => {
-                        setInvoiceSort(invoiceSort === 'desc' ? 'asc' : 'desc')
-                      }}
-                    >
+          {invoices.length > 0 && (
+            <main className="flex flex-row gap-4 items-center border rounded-lg">
+              <table className="min-w-full bg-white rounded-lg overflow-hidden">
+                <thead className="bg-gray-50 rounded-t-lg">
+                  <tr className="rounded-t-lg">
+                    <th className="px-6 py-3 text-left rounded-tl-lg">
+                      <button
+                        className="flex flex-row gap-2"
+                        onClick={() => {
+                          setInvoiceSort(
+                            invoiceSort === 'desc' ? 'asc' : 'desc'
+                          )
+                        }}
+                      >
+                        <Text
+                          size="xs"
+                          weight="medium"
+                          className="uppercase text-gray-500"
+                        >
+                          Date
+                        </Text>
+
+                        <svg
+                          width="16"
+                          height="16"
+                          viewBox="0 0 16 16"
+                          fill="none"
+                          xmlns="http://www.w3.org/2000/svg"
+                        >
+                          <path
+                            fillRule="evenodd"
+                            clipRule="evenodd"
+                            d="M8.00005 2.40002C8.21222 2.40002 8.41571 2.48431 8.56573 2.63434L10.9657 5.03434C11.2782 5.34676 11.2782 5.85329 10.9657 6.16571C10.6533 6.47813 10.1468 6.47813 9.83436 6.16571L8.00005 4.3314L6.16573 6.16571C5.85331 6.47813 5.34678 6.47813 5.03436 6.16571C4.72194 5.85329 4.72194 5.34676 5.03436 5.03434L7.43436 2.63434C7.58439 2.48431 7.78788 2.40002 8.00005 2.40002ZM5.03436 9.83434C5.34678 9.52192 5.85331 9.52192 6.16573 9.83434L8.00005 11.6687L9.83436 9.83434C10.1468 9.52192 10.6533 9.52192 10.9657 9.83434C11.2782 10.1468 11.2782 10.6533 10.9657 10.9657L8.56573 13.3657C8.25331 13.6781 7.74678 13.6781 7.43436 13.3657L5.03436 10.9657C4.72194 10.6533 4.72194 10.1468 5.03436 9.83434Z"
+                            fill="#9CA3AF"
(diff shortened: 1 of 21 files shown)


apps/console/app/routes/__layout/gnillib/index.tsx
```diff
@@ -74,6 +82,12 @@ type LoaderData = {
   }
   successToast?: string
   connectedEmails: DropdownSelectListItem[]
+  invoices: {
+    amount: number
+    timestamp: number
+    status: string | null
```
szkl, line 88: Is it possible to have a default value?

apps/console/app/services/billing/stripe.ts
```diff
@@ -77,10 +83,26 @@ export const updateCustomer = async ({
   return customer
 }
 
+export const accessCustomerPortal = async ({
+  customerID,
+  returnURL,
+}: CustomerPortalParams) => {
+  const stripeClient = new Stripe(STRIPE_API_SECRET, {
+    apiVersion: '2022-11-15',
```
szkl, line 91: It might be a good idea to make this an environment variable.

platform/address/src/jsonrpc/methods/sendReconciliationNotificationMethod.ts
```diff
@@ -0,0 +1,42 @@
+import { z } from 'zod'
+
+import { Context } from '../../context'
+import { ServicePlanType } from '@proofzero/types/account'
+import { ReconciliationNotificationType } from '../../../../email/src/jsonrpc/methods/sendReconciliationNotification'
```
szkl, line 5: `@proofzero/platform.client`?

platform/starbase/src/jsonrpc/methods/reconcileAppSubscriptions.ts
```diff
@@ -0,0 +1,104 @@
...
+        appURN: edge.dst.baseUrn,
+        devEmail,
+      })
+    }
+  }
+
+  const planApps = apps.filter((app) => app.appPlan === plan)
+  if (planApps.length > count) {
+    const targetApps = planApps
+      .sort((a, b) => +b.createdTimestamp! - +a.createdTimestamp!)
```
szkl, line 77: Is it possible to avoid `!`? `undefined` might not work well with subtraction.

apps/console/app/services/billing/stripe.ts
Cosmin-Parvulescu:
```
    export type LatestApiVersion = '2022-11-15';
    export type HttpAgent = Agent;
    export type HttpProtocol = 'http' | 'https';

    export interface StripeConfig {
      /**
       * This library's types only reflect the latest API version.
       *
       * We recommend upgrading your account's API Version to the latest version
       * if you wish to use TypeScript with this library.
       *
       * If you wish to remain on your account's default API version,
       * you may pass `null` or another version instead of the latest version,
       * and add a `@ts-ignore` comment here and anywhere the types differ between API versions.
       *
       * @docs https://stripe.com/docs/api/versioning
       */
      apiVersion: LatestApiVersion;
      ```
      
      If we put an env variable, we'd still need to cast it to the actual string value, not sure that's worth it or cleaner.

File: apps/console/app/routes/__layout/gnillib/index.tsx (hunk 1 above)
Comment (Cosmin-Parvulescu) [reply apps/console/app/routes/__layout/gnillib/index.tsx:88]: Hmm, it's currently `string | null` to cast nicely from Stripe client's types:
```
      /**
       * The status of the invoice, one of `draft`, `open`, `paid`, `uncollectible`, or `void`. [Learn more](https://stripe.com/docs/billing/invoices/workflow#workflow-overview)
       */
      status: Invoice.Status | null;
```

I guess there could be some default value, but we'd need to handle the casting from the stripe res, not sure if better:
```
      invoices = stripeInvoices.invoices.data.map((i) => ({
        amount: i.total / 100,
        timestamp: i.created * 1000,
        status: i.status, // Here
        url: i.hosted_invoice_url ?? undefined,
      }))
```

File: platform/starbase/src/jsonrpc/methods/reconcileAppSubscriptions.ts (hunk 4 above)
Comment (Cosmin-Parvulescu) [reply platform/starbase/src/jsonrpc/methods/reconcileAppSubscriptions.ts:77]: Due to undefined being a possibility, there's a condition (L53) when adding items to the `apps` list:
```
    if (appDetails.createdTimestamp != null) {
```
so we'd expect only items with a timestamp to be present in the `planApps`. I think it's a case of Typescript not picking it up, so the `!` is mostly to stop it from crying. 😁

platform/address/src/jsonrpc/methods/sendReconciliationNotificationMethod.ts
Cosmin-Parvulescu: Good catch! I extracted the enum into the types package as it spans both e-mail and address workers.

apps/console/app/services/billing/stripe.ts
szkl: I meant it for updating the value actually.

szkl (approved):

